Issue #4, opened 2019-06-21
Add Prometheus Exporter Deployment


HoustonPutman:
The Solr operator should auto-run a prometheus exporter with SolrClouds that generates metrics.

I would imagine that there would be some configuration added to the SolrCloud CRD, like tagging information for Prometheus discovery.

In general we first need an official build of the Prometheus Exporter in a Docker container.
This should probably be added in the official [docker-solr](https://github.com/docker-solr/docker-solr) repo. I will be creating an issue there to see if we can have builts of the prometheus exporter made automatically.

HoustonPutman:
Maybe this should be implemented as a separate CRD. So solr clouds/standalones can be monitored easily even if they themselves aren't run in kubernetes.

HoustonPutman:
Relies on docker-solr/docker-solr#222

PR #6, opened 2019-06-24
Solr metrics added, along with tests.

HoustonPutman:
Fixes: #4 

**Testing performed**
Added tests to make sure the correct kube objects are made for the prometheus exporter CRD.

Also tested that the necessary labels are present on the service, and that the service is pointing to the correct pods.


HoustonPutman:
I will do some additional tests next week, and merge if everything works out.

PR #27, opened 2019-10-03
Making java memory configurable

sepulworld:
Signed-off-by: Zane Williamson <zanew@zillow.com>

Describe your changes
Remove hardcoded memory allocation, and give ability to update jvm parameter.

Testing performed
Tested locally with solrcloud 8.2
Tested to verify present hardcoded value works as default

### DIFF
--- a/config/crds/solr_v1beta1_solrcloud.yaml
+++ b/config/crds/solr_v1beta1_solrcloud.yaml
@@ -98,6 +98,8 @@ spec:
                 tag:
                   type: string
               type: object
+            solrJavaMem:
+              type: string
             zookeeperRef:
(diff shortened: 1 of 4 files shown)


sepulworld:
@HoustonPutman  this patch doesn't contain the "kustomize" fix so Travis will fail to install dependencies. Perhaps skip Travis and merge? Or should we apply the kustomize fix found in https://github.com/bloomberg/solr-operator/pull/26/commits

HoustonPutman:
I'll try to merge #26 first then we can rebase this and merge it. Hopefully that works.

Using the master go branch for builds definitely adds some fun. Thanks for finding that fix.

Issue #46, opened 2019-11-04
Solr pods never become ready when creating a new SolrCloud


timterle:
**Describe the bug**
When creating a SolrCloud the example-solrcloud-0 pod never becomes ready.

**To Reproduce**
Create a simple SolrCloud example with operator args:
- -zk-operator=true
- -etcd-operator=false

In the logs of the pod example-solrcloud-0 this is found:

```
java.lang.IllegalArgumentException: A HostProvider may not be empty!
	at org.apache.zookeeper.client.StaticHostProvider.<init>(StaticHostProvider.java:88)
	at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:449)
	at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:383)
```

kubectl shows this:
`pod/example-solrcloud-0                             0/1     Running             0          3m`

The solr pod never becomes ready. When checking the pod it says ZK_HOST is "/". When watching the solr statefulset it looks like the pod template has a correct ZK_HOST. But currentRevision and updateRevision for the solr statefulset differ so it's likely that the first revision of the statefulset has an invalid ZK_HOST and the second revision has a correct one. But the pod is stuck with the first revision of the statefulset and can't recover. Kubernetes tries to restart the pod many times, but it's always the same error because ZK_HOST is still "/".

A workaround is to delete the example-solrcloud-0 pod, the installation will then continue and succeed.

sepulworld:
Ah interesting, I think I’ve seen this same race condition as well. Does it happen to you every single time you setup a new SolrCloud? I never dug into why, but your debugging seems to have found the root cause! Nice
    	
    	

    	Get Outlook for iOS
    
  




On Mon, Nov 4, 2019 at 7:55 AM -0800, "timterle" <notifications@github.com> wrote:












Describe the bug

When creating a SolrCloud the example-solrcloud-0 pod never becomes ready.


To Reproduce

Create a simple SolrCloud example with operator args:

-zk-operator=true
-etcd-operator=false



In the logs of the pod example-solrcloud-0 this is found:
java.lang.IllegalArgumentException: A HostProvider may not be empty!
	at org.apache.zookeeper.client.StaticHostProvider.<init>(StaticHostProvider.java:88)
	at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:449)
	at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:383)



kubectl shows this:

[...]

PR #47, opened 2019-11-04
Make the created statefulset have the new reconciled solr cloud status

timterle:
Signed-off-by: Tim Terlegård <tim.terlegard@infomaker.se>

*Issue number of the reported bug or feature request: #46*

**Describe your changes**
Changed the GenerateStatefulSet function to accept the reconciled zk status as an argument. The zk status may be incorrect in the solrCloud argument.

**Testing performed**
Tested this change in a custom operator. Solr pods start up fine.


timterle:
@HoustonPutman @sepulworld  Do you understand why the test fails? The code change in feature/invalid-zkhost-in-statefulset is tiny and should not make the test fail like this.

sepulworld:
It is probably flakey TravisCI environment. If the `make test` works locally, then probably a re-run of the failed job in TravisCI will complete successfully.

HoustonPutman:
this is a great change that I am actually running into now. Thanks for finding the issue.

Let me re-run the tests.

HoustonPutman:
Ok, so the tests aren't failing. It's the manifests generation, which is even weirder. I say let's just wait till #38 is merged, and then rebase this and see if the issues are fixed.

HoustonPutman:
@timterle have you given maintainers permissions to commit to your branch? I did a merge and wanted to push it, then squash into master. But it looks like I can't.

Also this is a good change to make, but I think that the issue that exposed this won't happen anymore, as the operator waits to create the stateful set till the individual node services have been created. But this is the right implementation anyways.

timterle:
@HoustonPutman You're right, I didn't give anyone any permissions. Looks like you managed to merge it anyways. 👍

Issue #48, opened 2019-11-05
Support creating Standard Collection Aliases


sepulworld:
**Is your feature request related to a problem? Please describe.**
CREATEALIAS registers the alias name with the names of one or more collections provided by the command.

Probably do this via a new CRD, "SolrCollectionAlias"

**Describe the solution you'd like**

```
apiVersion: solr.bloomberg.com/v1beta1
kind: SolrCollectionAlias
metadata:
  name: collection-alias-foo
spec:
  solrCloud: example
  collections:
    - example
    - example2
  aliasType: standard
```

sepulworld:
I would like to implement this, but will wait for or base my changes off of https://github.com/bloomberg/solr-operator/pull/38

sepulworld:
I have a functional prototype for this now. I based the work off of kubebuilder v2.

https://github.com/swarupdonepudi/solr-operator/pull/2

PR #57, opened 2019-12-03
adding solr collection alias crd

sepulworld:
Signed-off-by: Zane Williamson <zanew@zillow.com>

*Issue number of the reported bug or feature request: #48 *

**Describe your changes**
Adding ability for solr-operator to manage lifecycle of solr collection standard aliases.

**Testing performed**
Local development testing for adding and removing alias


README.md
```diff
@@ -293,7 +315,7 @@ spec:
                      Required to use the `ProvidedZookeeper.Zetcd` option within the Spec.
                      If _true_, then an Etcd Operator must be running for the cluster.
                      ( _true_ | _false_ , defaults to _false_)
-* **-ingress-base-url** If you desire to make solr externally addressable via ingresses, a base ingress domain is required.
+* **-ingress-base-domain** If you desire to make solr externally addressable via ingresses, a base ingress domain is required.
                         Solr Clouds will be created with ingress rules at `*.(ingress-base-url)`.
```
HoustonPutman: can you change this to `ingress-base-domain`?

controllers/util/collection_util.go
```diff
@@ -59,6 +60,31 @@ func CreateCollection(cloud string, collection string, numShards int64, replicat
 	return success, err
 }
 
+// CreateCollecttionAlias to request the creation of an alias to one or more collections
+func CreateCollecttionAlias(cloud string, alias string, aliasType string, collections []string, namespace string) (err error) {
```
HoustonPutman: type in "Collection" in both method and comment above.

example/test_solrcloud.yaml
```diff
@@ -17,12 +17,12 @@ spec:
   zookeeperRef:
     provided:
       zookeeper:
-        persistentVolumeClaimSpec:
+        persistence:
```
HoustonPutman: Need to add spec in there:
```
persistence:
  spec:
    storageClassName....
```

api/v1beta1/solrcollectionalias_types.go
```diff
@@ -0,0 +1,72 @@
...
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SolrCollectionAliasSpec defines the desired state of SolrCollectionAlias
+type SolrCollectionAliasSpec struct {
+	// A reference to the SolrCloud to create alias on
+	SolrCloud string `json:"solrCloud"`
```
HoustonPutman: We should eventually support clouds not running in kube, via a zookeeper connection string. But we also need to do this for the `SolrCollection` CRD.

This can be a feature added in the future.

controllers/solrcollectionalias_controller.go
```diff
@@ -0,0 +1,147 @@
...
+// SolrCollectionAliasReconciler reconciles a SolrCollectionAlias object
+type SolrCollectionAliasReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=solr.bloomberg.com,resources=solrcollectionaliases,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=solr.bloomberg.com,resources=solrcollectionaliases/status,verbs=get;update;patch
+
```
HoustonPutman: you should add rules to `get` SolrCloud/SolrCloudStatus

controllers/util/collection_util.go
sepulworld: bah! good catch

api/v1beta1/solrcollectionalias_types.go
sepulworld: I see, so in future make this optional and couple with another input that is a zk connection string. Perhaps add in a new issue?

HoustonPutman: I see it as adding a solrCloud field with lets you either specify a ZkConnectionString, URL or NamespacedName.

We can make it backwards compatible for a bit whenever we make the change. But we should make sure it's comprehensive and add it for both the Collection and Alias at that same time. I'll make an issue

Issue #20, opened 2019-09-17
Add a Helm Chart for deploying the Solr Operator


HoustonPutman:
This could make it easier for people to deploy and upgrade the Solr Operator to their clusters.

sepulworld:
I think since the CRDs are autogenerated it might make best sense to host the helm code and subsequent .tar.gz helm packages inside of this Github repo and enable the 'GitHub Pages' functionality on this repo to serve the Helm charts (see https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417 for how to enable GitHub Pages on a repo and allow it to be a repo for helm charts). We have done something similar with a chart we manage: https://github.com/sepulworld/dinghy-ping-helm-chart

It would be cool if Kubebuilder supported helm chart generation, but it presently doesn't do this directly. (some discussion on that topic here: https://github.com/kubernetes-sigs/kubebuilder/issues/387). But we could probably leverage Make and Kustomize to make it work for us I bet. 💯

kcmartin:
Just an FYI on this issue-- myself and @sepulworld have a WIP Helm chart PR on [Swarup's fork](https://github.com/swarupdonepudi/solr-operator/pull/3) of the Solr Operator

PR #60, opened 2019-12-10
solr-operator Helm chart

kcmartin:
Signed-off-by: Kristin Martin <kcmartin2@mac.com>

*Issue number of the reported bug or feature request: #20 

**Describe your changes**
Adding Helm chart to install solr-operator.
For now, this lives within the solr-operator repo.

**Testing performed**
I tested installation on a local dev cluster, and on one of our dev clusters in AWS

Tested installation and removal using both Helm 2 and Helm 3, see notes README regarding CRD handling in Helm 2 vs. 3

**Additional context**
@swarupdonepudi mentions we will need to add RBAC to the chart, so I would like to get feedback from @HoustonPutman and @sepulworld regarding this.

Thank you!


HoustonPutman:
Thanks for the addition kristin! I'm not very knowledgable about Helm at all, but I will try to learn as I review this. Sorry if this takes some time though. I'll probably have more than a few questions.

kcmartin:
Thanks @HoustonPutman! 

We also have @sepulworld and @swarupdonepudi on the team to assist with Helm-related questions.

I'll circle back with them about adding the RBAC rules.

One more note, It looks like the Travis build is failing due to `kustomize` being missing as a dependency, is that something I should add to https://github.com/bloomberg/solr-operator/blob/master/hack/install_dependencies.sh ?

HoustonPutman:
Yeah, for kustomize that's the right way to go.

I'd use the `install` method mentioned here: https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md#try-go

kcmartin:
I'm going to go ahead and close this PR, in favor of a new one that is cleaner and doesn't include weird merges :)

Issue #63, opened 2019-12-30
Ability to request custom template annoations for StatefulSet of SolrCloud


sepulworld:
**Is your feature request related to a problem? Please describe.**
We would like to apply kube2iam roles to pods in solrcloud statefulset. The role will allow access to s3 for backups. 

Snippet of what is needed
```
spec:
  template:
    metadata:
      annotations:
        iam.amazonaws.com/role: role-arn
```

**Describe the solution you'd like**

The ability to add any spec/template/metadata annotation for developers to request

HoustonPutman:
that sounds good. We should already have something like this, but I think just for labels, not annotations. And I could be wrong about that too...

Basically we need to enable a lot more customization of pod preferences

HoustonPutman:
`Spec.customSolrKubeOptions.podOptions` and `Spec.customSolrKubeOptions.statefulSetOptions` have been added.

Issue #55, opened 2019-11-20
More configuration options


meskallito:
Hi there,

We are considering using Bloomberg operator for our SOLR cluster but there are a few things we need to configure that the operator doesn't provide for us:

- Support of configuration for security contexts on stateful sets 
- Custom annotations on services and stateful sets
- Adding our data volumes to stateful sets
- Setting policy name on collections

HoustonPutman:
Hey @meskallito thanks for the input. We can definitely look at adding these things, they seem reasonable to me at a first-glance.

We are no longer blocked on #38 , so if you are interested in creating a PR with one or more of these added, we will be happy to look over and review it! And hopefully eventually we get all of these features added.

HoustonPutman:
`Spec.customSolrKubeOptions.podOptions` and `Spec.customSolrKubeOptions.statefulSetOptions` have been added.

This allows for custom:
- Security Context
- Volumes
- Annotations on statefulSets/pods

What we still need:
- Service annotations
- Setting policy name on collections.

HoustonPutman:
Service annotations have been added in `Spec.customSolrKubeOptions.commonServiceOptions`, `Spec.customSolrKubeOptions.headlessServiceOptions` and `Spec.customSolrKubeOptions.nodeServiceOptions`.

Last request is setting the policy name on a collection.

HoustonPutman:
Since the SolrCollection CRD has been removed, this issue can be closed. All features on the SolrCloud CRD have been added.

pareekdevanshu:
@HoustonPutman I was looking for an option to set securityContext for solr-cloud containers but couldnt find any option in the CRD to do that. I only see option to set securityContext for `sidecarContainers` and `initContainers`.
Am I missing something?

HoustonPutman:
You can set the podSecurityContext through `customSolrKubeOptions.podOptions.podSecurityContext`

pareekdevanshu:
@HoustonPutman Thanks for the reply, I wanted to `drop all capabilities in security context `  which can be done using securityContext :
```
securityContext:
  capabilities:
    drop:
      - ALL
     
``` 
can you please tell me how can we achieve the same using `customSolrKubeOptions.podOptions.podSecurityContext`? 
I couldnt find a way to mention the capabilities in `podSecurityContext`.

HoustonPutman:
Ahh it looks like you can only set capabilities on the container security context. If you make a separate issue we can try to get that added.

[...]

Issue #36, opened 2019-10-16
Enable configuring of hostname generation and ports


timterle:
**Is your feature request related to a problem? Please describe.**
We aim to run Solr in Amazon EKS (managed Kubernetes). We have microservices in an Amazon VPC that wants to communicate with Solr and Zookeeper running in a Kubernetes cluster that belongs to the same VPC. Amazon EKS exposes pod IPs in the VPC, but not service IPs. So our microservices (outside of Kubernetes but in the same VPC) can only talk to Solr pods, not Solr services. The Solr Operator can't easily be configured to enable this use-case.

**Describe the solution you'd like**
Possibility to configure template like expression for hostnames: {{ .Name }}.{{ .Namespace }}.ourdomain.com
Possibility to configure service port. It's now hardcoded to 80.

**Additional context**
By using ExternalDNS (https://github.com/kubernetes-incubator/external-dns) and annotating the Solr and Zookeeper headless services we can get Solr and Zookeeper pod IPs (VPC IPs, not Kubernetes internal IPs) registered in DNS (Route53). Our microservices can use these DNS records to talk to Solr pods.

As an example, let's say we create a solrcloud "example" with ingress-base-domain=ourdomain.com, namespace=default, ExternalDNS enabled, the result is then

hosts registered in zookeeper:
  default-example-solrcloud-0.ourdomain.com:80
  default-example-solrcloud-zookeeper-0.ourdomain.com:80
pods:
  example-solrcloud-0 8983
  example-solrcloud-zookeeper-0 8983
services:
  example-solrcloud-0 80
  example-solrcloud-zookeeper-0 80
dns:
  example-solrcloud-0.ourdomain.com
  example-solrcloud-zookeeper-0.ourdomain.com

With the help of ExternalDNS our own microservices can now use example-solrcloud-zookeeper-0.ourdomain.com:2181 as address to zookeeper. It will resolve to the Zookeeper pod IP in the VPC. However, the solrj client will get the state of the cluster from Zookeeper and it says all hosts are prefixed with "default-". So solrj tries to use the hostname default-example-solrcloud-0.ourdomain.com to talk to solr. But this host does not exist in DNS. If it was possible to configure the Solr Operator how the hostnames are generated it would be possible to skip the namespace and generate the same hostnames as ExternalDNS (example-solrcloud-0.ourdomain.com).

[...]

Issue #32, opened 2019-10-07
Support the Ability to tune GC_TUNE and SOLR_OPTS


sepulworld:
**Is your feature request related to a problem? Please describe.**
Ability to tune SOLR_OPTS and GC_TUNE, related docs
https://lucene.apache.org/solr/guide/8_1/taking-solr-to-production.html#memory-and-gc-settings
https://lucene.apache.org/solr/guide/8_1/taking-solr-to-production.html#override-settings-in-solrconfig-xml

**Describe the solution you'd like**
When going to production these values require tuning typically. Allow for the Solr-Operator SolrCloud CRD the ability to tune values passed to these ENV variables for Solr Cloud process to use.

Im starting work on this now. If you have any feedback or suggestions @HoustonPutman let me know! should be straightforward though.

timterle:
Hi, I'm also also interested in the ability to tweak the solr arguments.

We are thinking about creating our own solr image based on the official one and do some special work in the entrypoint based on some custom environment variables. So we'd like to be able to specify any environment variable.

Would it be possible to just have an "env" variable that can include any key/value mapping?

sepulworld:
Thanks for the feedback @timterle. Yes, this is probably possible, and would offer more flexibility in long run. I will take a look

HoustonPutman:
Support for custom Env variables was added in #74 !

PR #68, opened 2020-01-03
Add new customization API

HoustonPutman:
*Issue number of the reported bug or feature request: #63, #55, #36, #32*

**Describe your changes**
This is a PR to framework all of the new configuration options that we want to implement.

We can split further implementation into separate PRs, but I want to agree on API before any implementation is done.

Also the only customization supported currently is the SolrCloud object, not ZK, Zetcd, Solr Prometheus Exporter, etc. So we will also need separate PRs/Issues to deal with those. But I think that Solr should be the first we tackle it for.

**Testing performed**
Still need to write tests, and do implementation. We will likely need to write a very flexible testing framework so that all of these dozens of options can be tested completely.

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -1,0 +1,138 @@
+package v1beta1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"strings"
+)
+
+// StatefulSetOptions defines custom options for StatefulSets
+type StatefulSetOptions struct {
+	// Annotations to be added for the solrCloud Services.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Labels to be added for the solrCloud Services.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// PodOptions defines the common pod configuration for Pods, including when used
+// in deployments, stateful-sets, etc.
+type PodOptions struct {
+	// The scheduling constraints on pods.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+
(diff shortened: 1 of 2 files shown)


sepulworld:
This looks like a great start! I see you defined the data inputs, next step would be to introduce into the generator functions? Let me know if I can help out.

sepulworld:
We should add ImagePullSecret, example of current use "solrPrometheusExporter.Spec.Image.ImagePullSecret". Probably should be part of the customization API. Ill send a PR to your branch to consider.

sepulworld:
> We should add ImagePullSecret, example of current use "solrPrometheusExporter.Spec.Image.ImagePullSecret". Probably should be part of the customization API. Ill send a PR to your branch to consider.

Sorry, nevermind, I see it now in common_types.go

HoustonPutman:
Would love to get y'alls thoughts @meskallito and @timterle, as this would hopefully address the Issues that you raised.

HoustonPutman:
After confirming the options I will split the PR into separate PRs, one for each of the kube resource we are exposing options for (StatefulSet, pod, deployment, etc.) That way we can iterate on them independently, and everything isn't blocked if one feature is a little harder to do.

HoustonPutman:
So `Spec.customSolrKubeOptions.podOptions` and `Spec.customSolrKubeOptions.statefulSetOptions` have been added via #74 . Will hopefully look into configMap/Service/Ingress options next.

HoustonPutman:
The other resource options were added via #75.

Now we just need to work on the customized addressability.

HoustonPutman:
Closing as #95 has been opened to implement the last feature here.

PR #70, opened 2020-01-10
Helm chart for solr-operator-take 2

kcmartin:
*Issue number of the reported bug or feature request: #20  

**Describe your changes**
Adding Helm chart to install solr-operator--new PR, replaces #60 
For now, this lives within the solr-operator repo.

**Testing performed**
I tested installation for the previous PR on a local dev cluster, and on one of our dev clusters in AWS

Tested installation and removal using both Helm 2 and Helm 3, see notes README regarding CRD handling in Helm 2 vs. 3 -- need to expand on the README

**Additional context**
Added RBAC files, tested using Helm 2.12, 02/07/20


helm/solr-operator/values.yaml
```diff
@@ -0,0 +1,26 @@
+# Default values for solr-operator.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: bloomberg/solr-operator
+  tag: v0.2.1
```
HoustonPutman: It would be amazing if we could auto-update this on every build, like we do with the kustomize stuff here:
https://github.com/bloomberg/solr-operator/blob/master/Makefile#L57

I've tried to automate away any explicit references to a version. So that we can release without having to update versions.

helm/solr-operator/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart Solr Operator
+name: solr-operator
+version: 0.1.0
```
HoustonPutman: Should this match the version of the operator? Not sure how this is generally used.

helm/solr-operator/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+
+
+## Chart Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| useEtcdOperator | string | `"false"` |  |
```
HoustonPutman: I will reorder these and add some descriptions

helm/solr-operator/values.yaml
HoustonPutman: Honestly maybe we should just set this to `latest`?

helm/solr-operator/README.md
```diff
@@ -36,18 +36,25 @@ The command removes all the Kubernetes components associated with the chart and
...
-| useZkOperator | string | `"true"` |  |
\ No newline at end of file
+
+### Configuring the Solr Operator
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| ingressBaseDomain | string | `""` | If you have a base domain that points to your ingress controllers for this kubernetes cluster, you can provide this. SolrClouds will then begin to use ingresses that utilize this base domain. E.g. `solrcloud-test.<base.domain>` |
+| useZkOperator | string | `"true"` | This option enables the use of provided Zookeeper instances for SolrClouds |
+| useEtcdOperator | string | `"false"` | This option enables the use of provided Zetcd instances for SolrClouds |
```
kcmartin: Zetcd -> etcd ?

helm/solr-operator/values.yaml
kcmartin: I'm good with `latest`  if you don't think it has the capacity to cause any problems

helm/solr-operator/Chart.yaml
kcmartin: If you're pointing to latest, how would we determine the matching version?

kcmartin: Seems arbritrary? Here's a link on version & appVersion:
https://codefresh.io/docs/docs/new-helm/helm-best-practices/#chart-versions-and-appversions

kcmartin: Generally, I've incremented `version` with each Helm chart change, `appVersion` is optional, perhaps we leave it out?

helm/solr-operator/README.md
HoustonPutman: So Zetcd is a cluster of Etcd, and a deployment of Zetcd. So you need the Etcd Operator running in order to use Zetcd (A small proxy in front of etcd).

helm/solr-operator/values.yaml
HoustonPutman: I agree that it should be the highest version, and I would love to have some automated scripts to go and update the versions throughout the project everytime a release is made. Until then I'd say we should probably use latest.

It's at least a small improvement we can make in the future. Getting this merged is more important 😃

helm/solr-operator/Chart.yaml
HoustonPutman: Yeah, we won't change the inputs to the solr operator very often I think. So keeping this as 0.1.0 is probably fine.

Issue #53, opened 2019-11-11
AWS EKS Multi AZ


AceHack:
I cannot find a way to make this operator spread across AZs in AWS, any suggestions would be welcome.

More info
https://github.com/bitnami/charts/issues/927
https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

HoustonPutman:
Hey @AceHack , sorry for the delayed response. Things have been pretty crazy, and we have been blocked on the Kubebuilder v2 migration. Finally able to get into these other issues.

In general, Solr is relatively easy to run via multiple kubernetes clusters. What you need to do is make sure that you are using 1 zookeeper cluster for the solr clouds running in all kube clusters, so that they all share the same zookeeper connection information in their SolrCloud Specs. The data volumes don't have to be linked at all, since each Solr node manages it's data independently of other Solr Nodes.

We use different base ingress URLs to differentiate SolrCloud nodes in each kubernetes cluster.

So the solr-operator would be started in Kube cluster A with `--base-ingress-domain a.test.com` and in Kube cluster B with `--base-ingress-domain b.test.com`. Then when you create a SolrCloud `example` in each kube cluster (both connecting to the exact same zookeeper connection information), the nodes will register under their cluster-specific domain names. You just need to make sure that your DNS rules are set up correctly to route `a.test.com` and `b.test.com` to the correct ingress controllers.

We are hopefully going to be able to support more addressability options in the near future, such as ExternalDNS, and give you more control over how things are created/named.

AceHack:
We don't want to run this in multiple clusters, we have 1 kubernetes cluster that spans 3 AZs.  Currently, we have 3 different stateful sets (1 per AZ) so that Solr can be AZ aware in its autoscaling policies.  It also used to be a nightmare with PVs/PVCs until the `WaitForFirstConsumer` `volumeBindingMode` for StorageClasses came out.  To really support multi-AZ, I should be able to set a solr autoscaling policy that can reference the different AZs within the cluster.

AceHack:
I would really like to set a policy that looks something like

`{"replica": "#EQUAL", "shard": "#EACH", "sysprop.zone": "#ANY"}
`
or 

`{"replica": "#EQUAL", "shard": "#EACH", "sysprop.zone": ["us-east-1a", "us-east-1b", "us-east-1c"]}
`

AceHack:
More info
[...]

PR #80, opened 2020-02-20
Adding autoscaling policy configuration for Collection.

### LINKED ISSUE #52: Allow setting autoscaling policy
When creating a collection I would like to be able to create and associate an autoscaling policy.

https://lucene.apache.org/solr/guide/7_7/solrcloud-autoscaling-policy-preferences.html

### LINKED ISSUE #55: More configuration options

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #55, Resolves #52*

**Describe your changes**
Allowing for the setting of collection autoscaling policies via the Collection CRD.


AceHack:
Will I be able to target Avialiabily Zones in AWS EKS as part of the policy?

AceHack:
I would really like to set a policy that looks like

`{"replica": "#EQUAL", "shard": "#EACH", "sysprop.zone": "#ANY"}
`
or

`{"replica": "#EQUAL", "shard": "#EACH", "sysprop.zone": ["us-east-1a", "us-east-1b", "us-east-1c"]}
`

AceHack:
More info
https://lucene.472066.n3.nabble.com/Rule-based-replication-or-sharding-td4407125.html#a4407660

sepulworld:
This sounds great! We deal with really slow scale ups because new nodes don’t land in the right AZ where EBS is. 

“where "az" is a sysprop while starting each solr instance ( -Daz=us-east-1 )”
Sounds like the stateful set needs to pass a unique startup property to the instance depending on if it is in AWS and what AZ it is in. Might also be similar for Azure and GCP.
    	
    	

    	Get Outlook for iOS
    
  




On Fri, Feb 21, 2020 at 7:30 AM -0800, "Aaron Stainback" <notifications@github.com> wrote:












More info

https://lucene.472066.n3.nabble.com/Rule-based-replication-or-sharding-td4407125.html#a4407660

HoustonPutman:
Yeah, I've thought about that for a while. Maybe we can add Kubernetes Node labels to our Solr nodes somehow? 

How do y'all currently get solr to run across these availability zones? I'm assuming you have Kubernetes Node labels that describe which AZ each one is in, and you have Affinity rules setup to spread the solrcloud across them?

HoustonPutman:
Looks like the adding Node Labels as Pod environment variables has been discussed for a while and nothing has been done about it yet. https://github.com/kubernetes/kubernetes/issues/40610

That's unfortunate.

AceHack:
We've currently just created 3 separate stateful sets and each one is tied to one AZ and adds the appropriate zone property to the solr startup.  It's possible to read your AZ in an init container or as part of the pod startup, you'll have to call into the K8s APIs to get it.

sepulworld:
Another option we will be exploring...

https://github.com/kubernetes/autoscaler/blob/d4a6664b389b052293de8f2c0d16839564911e1f/cluster-autoscaler/FAQ.md#i-have-a-couple-of-pending-pods-but-there-was-no-scale-up

basically, we will move over to 3 ASGs, 1 per AZ (what @AceHack mentions above), and let the one statefulset generate by Solr-operator do its thing. Cluster-autoscaler will keep a balance between AZs and should keep us with pods that come online fast without PV toleration issues. Ill follow up with our findings after we test this out.

HoustonPutman:
Thanks for the link @sepulworld. Let's move this discussion over to #53, unless y'all think that we should block this PR until we can figure out a way to do this availability zone thing.

I think that the work for that is very important, but it could be done in separate PRs from this.

janhoy:
Please be aware of https://cwiki.apache.org/confluence/display/SOLR/SIP-8+Autoscaling+policy+engine+V2 in this work - autoscaling may be re-built for solr 9

HoustonPutman:
Thanks for linking this! It's definitely something to keep in mind. Especially if the operator/kubernetes could help handle the external orchestration.

HoustonPutman:
Closing as Autoscaling has been deprecated in recent releases of Solr.

PR #90, opened 2020-03-05
Adding Kubernetes Node labels to Solr nodes. (WIP)

### LINKED ISSUE #53: AWS EKS Multi AZ

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #53* 

**Describe your changes**
This PR adds Kubernetes Node labels as sysProps for Solr Nodes, thus allowing for them to be used within Solr autoscaling policies.

Kubernetes Node labels are added in the following steps:

1. In an `initContainer`, the Kubernetes API is queried to get the information for the node the pod resides on.
2. `jq` is used to parse this information and generate a sh script to update `$SOLR_OPTS` to contain `-d<kube-node-label>=<value>`.
3. This bash file is shared via a volume between the initContainer and the solr container.
4. Solr is provided this sh script via  the `$SOLR_INCLUDES` envVar. This is an option that [Solr reads when it starts up](https://github.com/apache/lucene-solr/blob/master/solr/bin/solr#L82).
5. When Solr starts, the SOLR_OPTS will contain the `-D` options that specify all of the Kubernetes node labels.

**Testing performed**
Tested locally.

**This is a WIP and the feature has not been finalized yet.**


AceHack:
Any update?

HoustonPutman:
No update other than it worked in my tests. We probably want to add this as an optional feature, at least at first. I can probably do some work on it next week or the next one.

I wanted to get some feedback before moving forward with it

HoustonPutman:
I'm going to try and get #95 merged in, then cut the `v0.2.4` release. After that I'll try to get this merged, as it's quite a large change to how Solr is "run".

janhoy:
In k8s v1.35 pods automatically inherit some topology labels https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#pod-topology-labels and in such an environment the operator won't need to do anything other than pulling the labels into env.variables with downward API.

So we could choose to make the operator sniff the k8s version, and if >=1.35, prepare env-vars (which implicitly also sets sysprop). Then document that users >=1.35 can benefit from this, while users on older k8s versions will need a workaround, which is not provided by the operator?

PR #93, opened 2020-03-19
Upgrade Zookeeper Operator default version to 0.2.6

HoustonPutman:
**Describe your changes**
Making the version of the example zookeeper operator and the default zookeeper version the same.


HoustonPutman:
@timterle is there anywhere else we need to update?

timterle:
@HoustonPutman There is a dependency in go.mod and go.sum on v0.2.5-rc0, maybe that should be changed as well? Otherwise I think this is enough, although I haven't built and tested this to confirm.

I have tested changing from emccorp/zookeeper to pravega/zookeeper by overriding the image in the config and it works when I run a simple deployment in Amazon EKS. But when I try a simple example in "Docker for Mac" this change causes the second zookeeper pod never to come up successfully. It exits with an error "java.lang.RuntimeException: My id 2 not in the peer list"

sepulworld:
Definitely need this. We ran into a lot of issues HA testing out zk-operator with emccorp/zookeeper only to then find out it shouldn't be used according to reading through issue tracker on pravega/zookeeper-operator

HoustonPutman:
I'm going to do some testing with this tomorrow, and see if I can get this merged.

HoustonPutman:
Looks like `v0.2.6`, which uses gomodules. Might as well see if that one works too, keep this as up to date as possible.

HoustonPutman:
@timterle as a bonus with `v0.2.6`, you will be able to use the ExternalDNS with your zookeeper ensemble as well. https://github.com/pravega/zookeeper-operator/releases/tag/v0.2.6-rc0

HoustonPutman:
@timterle I have tested this on docker-for-mac and it works well for me, including scaling up and down the zookeeper cluster through the solrcloud CRD

PR #95, opened 2020-03-24
Adding custom addressability options to the SolrCloud spec.

### LINKED ISSUE #36: Enable configuring of hostname generation and ports

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #36*

**Describe your changes**
Added the following addressability options for SolrCloud. All changes are backwards compatible and will work with the `ingressBaseDomain` option for the solr-operator.

Eventually we will remove the option from the operator as a whole and require use of the following options:

```golang
type SolrAddressabilityOptions struct {
	// External defines the way in which this SolrCloud nodes should be made addressable externally, from outside the Kubernetes cluster.
	// If none is provided, the Solr Cloud will not be made addressable externally.
	// +optional
	External *ExternalAddressability `json:"external,omitempty"`

	// PodPort defines the port to have the Solr Pod listen on.
	// Defaults to 8983
	// +optional
	PodPort int `json:"podPort,omitempty"`

	// CommonServicePort defines the port to have the common Solr service listen on.
	// Defaults to 80
	// +optional
	CommonServicePort int `json:"commonServicePort,omitempty"`

	// KubeDomain allows for the specification of an override of the default "cluster.local" Kubernetes cluster domain.
	// Only use this option if the Kubernetes cluster has been setup with a custom domain.
	// +optional
	KubeDomain string `json:"kubeDomain,omitempty"`
}

// ExternalAddressability defines the config for making Solr services available externally to kubernetes.
// Be careful when using LoadBalanced and includeNodes, as many IP addresses could be created if you are running many large solrClouds.
type ExternalAddressability struct {
	// The way in which this SolrCloud's service(s) should be made addressable externally.
	Method ExternalAddressabilityMethod `json:"method"`

	// Use the external address to advertise the SolrNode, defaults to false.
	//
	// If false, the external address will be available, however Solr (and clients using the CloudSolrClient in SolrJ) will only be aware of the internal URLs.
	// If true, Solr will startup with the hostname of the external address.
	//
	// NOTE: This option cannot be true when hideNodes is set to true. So it will be auto-set to false if that is the case.
	//
	// Deprecation warning: When an ingress-base-domain is passed in to the operator, this value defaults to true.
	// +optional
	UseExternalAddress bool `json:"useExternalAddress"`

	// Do not expose the common Solr service externally. This affects a single service.
	// Defaults to false.
	// +optional
	HideCommon bool `json:"hideCommon,omitempty"`

	// Do not expose each of the Solr Node services externally.
	// The number of services this affects could range from 1 (a headless service for ExternalDNS) to the number of Solr pods your cloud contains (individual node services for Ingress/LoadBalancer).
	// Defaults to false.
	// +optional
	HideNodes bool `json:"hideNodes,omitempty"`

	// Override the domainName provided as startup parameters to the operator, used by ingresses and externalDNS.
	// The common and/or node services will be addressable by unique names under the given domain.
	// e.g. default-example-solrcloud.given.domain.name.com
	//
	// This options will be required for the Ingress and ExternalDNS methods once the ingressBaseDomain startup parameter is removed.
	//
	// For the LoadBalancer method, this field is optional and will only be used when useExternalAddress=true.
	// If used with the LoadBalancer method, you will need DNS routing to the LoadBalancer IP address through the url template given above.
	// +optional
	DomainName string `json:"domainName,omitempty"`

	// Provide additional domainNames that the Ingress or ExternalDNS should listen on.
	// This option is ignored with the LoadBalancer method.
	// +optional
	AdditionalDomainNames []string `json:"additionalDomains,omitempty"`

	// NodePortOverride defines the port to have all Solr node service(s) listen on and advertise itself as if advertising through an Ingress or LoadBalancer.
	// This overrides the default usage of the podPort.
	//
	// This is option is only used when HideNodes=false, otherwise the the port each Solr Node will advertise itself with the podPort.
	// This option is also unavailable with the ExternalDNS method.
	//
	// If using method=Ingress, your ingress controller is required to listen on this port.
	// If your ingress controller is not listening on the podPort, then this option is required for solr to be addressable via an Ingress.
	//
	// Defaults to 80 if HideNodes=false and method=Ingress, otherwise this is optional.
	// TODO: Should this be defaulted idk?
	// +optional
	NodePortOverride int `json:"nodePortOverride,omitempty"`
}


// ExternalAddressability is a string enumeration type that enumerates
// all possible ways that a SolrCloud can be made addressable external to the kubernetes cluster.
// +kubebuilder:validation:Enum=Ingress;ExternalDNS
type ExternalAddressabilityMethod string

const (
	// Use an ingress to make the Solr service(s) externally addressable
	Ingress ExternalAddressabilityMethod = "Ingress"

	// Use ExternalDNS to make the Solr service(s) externally addressable
	ExternalDNS ExternalAddressabilityMethod = "ExternalDNS"
)
```

**Testing performed**
There are currently unit tests for all of these features. Not all of them have been tested in a live cluster however.


sepulworld:
Any ETA on this? We would love to use / test it out!

Recently had to disable ingress on solr-operator because it would configure zk solrcloud host entries using the ingress fqdns which lead to performance issues when k8s solr client apps made requests using zk autoconfig lookups and used ingress to talk to solrcloud replicas. So, we have been setting up ingress outside of operator ever since.

sepulworld:
If you want to merge this first, I can rebase my latest MR (Since you handle port changes way better here! :) )

bsankara:
Hey @sepulworld i'm looking to get through this as soon as possible -- hopefully this week.

sepulworld:
Thanks @bsankara!

bsankara:
This looks great, thanks @HoustonPutman

PR #99, opened 2020-03-25
Fixed usage of empty routerField and made routerName an enum.

### LINKED ISSUE #97: Invalid router field is set on collection if none is specified in solrcollection yaml file
**Describe the bug**
When creating a collection and not specifying routerField it's set to the empty string which prevents anything from being indexed.

**To Reproduce**

```
apiVersion: solr.bloomberg.com/v1beta1
kind: SolrCollection
metadata:
  name: example-col
spec:
  solrCloud: example
  collection: example-col
  routerName: compositeId
  autoAddReplicas: false
  numShards: 1
  replicationFactor: 1
  maxShardsPerNode: 1
```
If I apply the above config I will not be able to index anything in the collection as router.field will be set to empty string and Solr will throw "SolrException: No value for :. Unable to identify shard"

**Expected behavior**
router.field should not be set at all if it's missing in the config

HoustonPutman:
*Issue number of the reported bug or feature request: Fixes #97*

**Describe your changes**
`routerField` is no longer added to the Solr Collections API call if it is empty.
Also changed `routerName` to be an enum, as there are only 2 possible options for this field.

### DIFF
--- a/api/v1beta1/solrcollection_types.go
+++ b/api/v1beta1/solrcollection_types.go
@@ -33,7 +33,7 @@ type SolrCollectionSpec struct {
 
 	// The router name that will be used. The router defines how documents will be distributed
 	// +optional
-	RouterName string `json:"routerName,omitempty"`
+	RouterName CollectionRouterName `json:"routerName,omitempty"`
 
 	// If this parameter is specified, the router will look at the value of the field in an input document
 	// to compute the hash and identify a shard instead of looking at the uniqueKey field.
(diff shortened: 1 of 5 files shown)


HoustonPutman:
Want to try this out @timterle ?

timterle:
@HoustonPutman I tested it now and can confirm that it's working

HoustonPutman:
Thanks for testing it out! I really appreciate it

Issue #77, opened 2020-02-20
Better documentation of each CRD and it's options


HoustonPutman:
Some CRDs for this operator, especially the [SolrCloud CRD](https://github.com/bloomberg/solr-operator/blob/master/config/crd/bases/solr.bloomberg.com_solrclouds.yaml), are getting extrememly large. This is because we are importing options from base kubernetes objects, which expand out to be extremely large.

Because these files are much to big to parse, we should have documentation for each CRD, explaining at least the basics of using each one. Eventually we can hopefully start to use links to the other types of resources, so they don't have to be explicitly listed in our CRDs. But for now I think the solution is just to add better documentation.

PR #110, opened 2020-04-27
Separate documentation from the README

### LINKED ISSUE #105: Better turorial for getting started
I got stuck several times in setting up the operator, as the README goes straight into building from source and it is a bit unclear what happens.

The first steps in README should not be building it from source but installing with as few steps as possible, whether using helm or directly with `kubectl apply` some combined yaml that gets everything setup (like for dependencies)

So when testing this out for myself, I created kind of a tutorial that is what I'd like to have seen when visiting the repo myself. It takes you through setting up your environment (Mac only) with ingress controller, k8s dashboard and Solr install, scale, upgrade and delete. The tutorial is in this gist, feel free to steal it or parts of it into a Wiki page or something in this repo.

https://gist.github.com/19bd70ab4ecbcc35eeaecfbc63fbbe1c

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #105 and relates to #77*

**Describe your changes**
I have created a documentation folder and split out a lot of the content from the top-level README into individual pages. I may have missed some stuff, but I think the majority has been put into better places.

I have also included the awesome tutorial provided by @janhoy. I tweaked this a bit, so I would definitely love additional comments.

The sections for each resource (SolrCloud, SolrCollection, etc) can definitely be improved, but I wanted to get this started first. Once we have the better layout, we can work on improving each section individually.

I would love as much feedback on this as possible before merging in, as it's a pretty dramatic change.


docs/local_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+	https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+
+# Install Docker Desktop for Mac (use edge version to get latest k8s)
+brew cask install docker-edge
+
+# Enable Kubernetes in Docker Settings, or run the command below:
+sed -i -e 's/"kubernetesEnabled" : false/"kubernetesEnabled" : true/g' \
+    ~/Library/Group\ Containers/group.com.docker/settings.json
+
+# Start Docker for mac from Finter, or run the command below
```
janhoy, line 29: Finder

docs/local_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+brew install helm watch
+```
+
+## Install an Ingress Controller
+
+Kubernetes services are by default only accessible from within the k8s cluster. To make them adressable from our laptop, we'll add an ingress controller
+
+```bash
+# Install the nginx ingress controller
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
```
janhoy, line 42: These urls are 404 now. I think they both can be replaced with this one instead https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud/deploy.yaml

docs/local_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+NAME                                       DESIRED   CURRENT   READY   AGE
+replicaset.apps/solr-operator-8449d4d96f   1         1         1       2d1h
+replicaset.apps/zk-operator-674676769c     1         1         1       49d
+```
+
+After inspecting the status of you Kube cluster, you should see a deployment for the Solr Operator as well as the Zookeeper Operator.
+
+## Start an example Solr cluster
+
+To start a Solr cluster, we will create a descriptor for the Solr Operator that will tell it what version of solr to install, and how many nodes, with how much memory etc.
```
janhoy, line 94: Not sure if 'descriptor' is the correct k8s term here?

docs/local_tutorial.md
HoustonPutman: yep, checked their documentation and that looks right! Good catch

docs/local_tutorial.md
```diff
@@ -0,0 +1,233 @@
+# Solr on Kubernetes on local Mac
```
nabadger, line 1: I commented on slack that I think this should be a more generalized example (i.e not specific to mac).

That would perhaps mean using minibike/k3d/KinD instead...

HoustonPutman: Would definitely love to have documentations for all of those options!

Are you using any of those other providers? If so would you be willing to help out with those steps?

janhoy: I have continued to develop the initial tutorial into a lab for a training course. For that I added instructions for Windows (which is a big hassle tbh). Feel free to copy/paste something from here: https://gist.github.com/4bc3398822b8b68ec94c60a05474e409

Since it is not mac-only I do not assume brew or anything, but have included choco commands. Powershell / CMD is a mess, so I added separate examples for them. Can you imagine that Powershell has a default alias for `curl` to some internal cmdled, even if you have the real curl.exe on path. ;(

HoustonPutman: That's great @janhoy!

I really wish there was a way in markdown to select an environment at the top and have that determine which example you saw. That would make the experience a whole lot cleaner.

HoustonPutman:
Ok so since there are no issues with the content currently, I will merge this now. But I will also open another issue to support non-docker-for-mac use cases as well. It'll be easier to iterate on that as a smaller PR.

PR #113, opened 2020-05-08
Adding hosted helm chart with deployment script.

HoustonPutman:
This is a PR to host a helm repo in the github pages of this github repository (bloomberg.github.io/solr-operator/charts). 

I still need to fix the documentation in the getting started, and add documentation around doing releases. But I think the logic should be there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ clean:
 mod-tidy:
 	export GO111MODULE=on; go mod tidy
 
+release:
+	VERSION=${VERSION} bash hack/setup_release.sh
+
 ###
 # Building
 ###
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ clean:
 mod-tidy:
 	export GO111MODULE=on; go mod tidy
 
-release:
+release: manifests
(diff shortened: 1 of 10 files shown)


HoustonPutman:
Tagging @kcmartin & @pawalt for reviews!

sepulworld:
This is awesome! We will definitely use the helm repo here!

Makefile
```diff
@@ -32,6 +32,9 @@ clean:
 mod-tidy:
 	export GO111MODULE=on; go mod tidy
 
+release:
```
pawalt, line 35: Maybe `helm-release` or `helm-publish` or something like that? Just don't want this to be confused with actually cutting a new release.

HoustonPutman: Added some docs around creating releases to make sure I (and others) don't miss anything in the future. I want this task to auto-create the release at some point...

Actually maybe I'll script out the necessary git commands.

pawalt: Gotcha that totally makes sense. Releases would probably be pretty easy to script out with the github cli, but that's a different PR.

Issue #126, opened 2020-07-17
Make Liveness and Readiness Checks for Solrcloud Statefulset Configurable


sepulworld:
**Is your feature request related to a problem? Please describe.**
Right now the liveness and readiness checks are hardcoded. I need some flexibility here when running backups to EFS on AWS. It generates high CPU IO Wait and I need more forgiving thresholds.

**Describe the solution you'd like**
Add to the solrcloud spec interface the different parameters that are currently hardcoded, set defaults to what they are now

sepulworld:
I can work on this one if you want to assign to me.

I think the best spec interface for these settings will be in https://github.com/bloomberg/solr-operator/blob/master/api/v1beta1/solrcloud_types.go#L66 @HoustonPutman sound good to you?

HoustonPutman:
Hey @sepulworld, I think this would be a great change! My only suggestion would be that you make a section in the top level SolrCloudSpec that you put these options under. I think that's what you meant, just wanted to make sure :)

Also @bsankara is the only one who can assign now, since I'm no longer at Bloomberg. He's a great replacement for me though!

HoustonPutman:
@sepulworld Anything else to do here, or is it ok to close?

sepulworld:
Yep! All working fine now 👍

PR #127, opened 2020-07-21
Adding configurable liveness and readiness checks

sepulworld:
Signed-off-by: Zane Williamson <zanew@zillowgroup.com>

*Issue number of the reported bug or feature request: #126 *

**Describe your changes**
Adding SolrCloudSpec Liveness and Readiness fields for configuration options. Kept existing values as defaults.

**Testing performed**
Local development and dev environment testing looks good.


HoustonPutman:
Hey @sepulworld . I think this is great, and would be very useful.

First easy thought, we should include startup probes as well.

I think my biggest thought is whether the whole default probe information should live in the CRD. I think it would be great to provide options for overriding the defaults in `spec.cutomSolrKubeOptions.podOptions.<readiness,startup,liveness>Probe`. But those overrides would be the only things stored. Then the overrides can be used in `solr_util`, which contains the default implementations.

That way when you run `kubectl describe solrCloud` it's a lot less verbose and it's easier to change defaults down the line.

The only issue I see with this approach (the one I described above) is that it's hard to specify that you want to not include a probe.... But maybe that's something that we don't allow... Maybe you have to have these probes.

I'd love to hear your thoughts on that.

(Also before we merge we should include it in at least a few of the unit tests)

sepulworld:
> Hey @sepulworld . I think this is great, and would be very useful.
> 
> First easy thought, we should include startup probes as well.
> 
> I think my biggest thought is whether the whole default probe information should live in the CRD. I think it would be great to provide options for overriding the defaults in `spec.cutomSolrKubeOptions.podOptions.<readiness,startup,liveness>Probe`. But those overrides would be the only things stored. Then the overrides can be used in `solr_util`, which contains the default implementations.
> 
> That way when you run `kubectl describe solrCloud` it's a lot less verbose and it's easier to change defaults down the line.
> 
> The only issue I see with this approach (the one I described above) is that it's hard to specify that you want to not include a probe.... But maybe that's something that we don't allow... Maybe you have to have these probes.
> 
> I'd love to hear your thoughts on that.
> 
> (Also before we merge we should include it in at least a few of the unit tests)

I see what you mean! Yeah, the full Probe into the CRD does inflate it. Let me try to put a default into solr_util and if user overrides with `spec.cutomSolrKubeOptions.podOptions.<readiness,startup,liveness>Probe` one of more, then do so.

I'll look into startup Probe as well!

sepulworld:
I took a look at the output of `kubectl describe solrcloud` with the current implementation.  Output of a solrcloud where I made adjustments to liveness, readiness....

```  Liveness:
    Failure Threshold:  3
    Http Get:
      Path:                 /solr/admin/info/system
      Port:                 8983
      Scheme:               HTTP
    Initial Delay Seconds:  20
    Period Seconds:         10
    Success Threshold:      1
    Timeout Seconds:        1
  Readiness:
    Failure Threshold:  3
    Http Get:
      Path:                 /solr/admin/info/system
      Port:                 8983
      Scheme:               HTTP
    Initial Delay Seconds:  15
    Period Seconds:         5
    Success Threshold:      1
    Timeout Seconds:        1
```

doesn't seem too overly verbose to me. Ill rework still and will compare.

api/v1beta1/solrcloud_types.go
```diff
@@ -221,7 +219,7 @@ func (spec *SolrCloudSpec) withDefaults(ingressBaseDomain string) (changed bool)
 			HTTPGet: &corev1.HTTPGetAction{
 				Scheme: corev1.URISchemeHTTP,
 				Path:   "/solr/admin/info/system",
-				Port:   intstr.FromInt(SolrClientPort),
+				Port:   intstr.FromInt(spec.SolrAddressability.CommonServicePort),
```
HoustonPutman, line 222: You want these to be PodPort

sepulworld: Good catch, ty!

sepulworld:
@HoustonPutman moved things around a bit. I also added the startupProbe, still need to test since this is only alpha in 1.16 and doesn't come with statefulSet unless running newer versions for K8s. Not sure if we can feature gate this one or not. But I found operator handles statefulSet generation fine even without startupProbe available in k8s api (say if you are running k8s 1.15)

I added a couple unit tests, and did some local testing with changes.

I can still look into removing corev1.Probe interface usage and go with something more generic for interface input. stand by for that

sepulworld:
Ha, maybe that was me on those other examples using k8s interfaces. Im down to not use them, I understand why when looking at CRD yaml outputs :)

sepulworld:
Besides deciding on interface adjustments, I tested startupProbe interface and it works as expected if provided by user

```  
customSolrKubeOptions:
    podOptions:
      livenessProbe:
        initialDelaySeconds: 20
        periodSeconds: 3
        successThreshold: 1
        timeoutSeconds: 2
        failureThreshold: 9
      readinessProbe:
        initialDelaySeconds: 21
        periodSeconds: 6
        successThreshold: 1
        timeoutSeconds: 3
        failureThreshold: 5
      startupProbe:
        exec:
          command:
          - "sleep"
          - "1"
        periodSeconds: 10
        failureThreshold: 30
```

Note for reference: startupProbe doesn't show up in `kubectl describe <pod>` but will be in `kubectl get <pod> -o yaml` in requested by user.

controllers/util/solr_util.go
```diff
@@ -353,6 +353,82 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 		if customPodOptions.NodeSelector != nil {
 			stateful.Spec.Template.Spec.NodeSelector = customPodOptions.NodeSelector
 		}
+
+		if customPodOptions.LivenessProbe.InitialDelaySeconds != 0 {
```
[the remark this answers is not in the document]
sepulworld: Good feedback! Ok, I created a fillProbe function that we can use anytime we detect user provides customPodOptions for any of the type of probe. And we will still default to generating our current Liveness and Readiness probes. Unit tests pass and I vetted out locally with a few different scenarios.

HoustonPutman:
👍 This looks good to me.

HoustonPutman:
@bsankara might ask you to squash into one commit. But I'll let him take the final call on that.

sepulworld:
Thanks @HoustonPutman and @bsankara! I squashed down to my commit and a rebase with master (since I opened this up prior to some merged in changes). I can close and reopen a new PR if you just want 1 commit w/o master rebase. Please let me know.

bsankara:
This looks great, thanks @sepulworld

sepulworld:
Great! Can we add this to a 0.2.6 release @bsankara? Thanks again!

ziodave:
There's a nice little script in the Solr docker image, `/opt/docker-solr/scripts/wait-for-solr.sh`:

This is how I configured the `startupProbe`:

```
      startupProbe:
        exec:
          command: 
            - /opt/docker-solr/scripts/wait-for-solr.sh
        periodSeconds: 10
        timeoutSeconds: 60
```

PR #133, opened 2020-08-17
Introduce managed SolrCloud update strategy

### LINKED ISSUE #66: Complications around readinessChecks and pod restarts
There is a chicken/egg problem with restarting SolrCloud pods.

- We need to be able to connect to a pod in order to do any necessary recovery of replicas contained on that Solr Node. Therefore the Service for that node needs to point to the pod.
- We need to ensure that a rolling restart of the StatefulSet does not start on the next pod until all replicas are healthy in the pod. The only way we can do that through the provided statefulSet mechanisms is through a robust `readinessCheck`.

So if we want both of these to be true, then the SolrNode services (both individual node services, and the headless service) need to route to pods regardless of whether that node is _ready_ or not. This can be achieved by using the following option:

```yaml
Service:
  Spec:
    PublishNotReadyAddresses: true
```

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #66*

**Describe your changes**
There has been much discussion around the way in which rolling upgrades should occur for Solr within Kubernetes.

Currently the operator lets the statefulset controller manage upgrades, in a rolling one-by-one strategy.
In the ticket linked above, there was discussion around the addition of a parameter in the healthcheck handler of solr to make sure that all Solr cores in a pod are active before issuing a restart command. This handler would make the default statefulset rolling upgrade strategy safer for SolrClouds, but it doesn't actually solve the issue of optimal ordering of pods for restarts or batching all pods that can be restarted at the same time.

This PR gives the Solr Operator the control over which pods are deleted and when. The operator sets the StatefulSet upgrade policy to be `OnDelete` and will choose a set of pods to be deleted at any given time when there are pods in a solrcloud statefulset that are not up to date. Currently the ordering for upgrading pods is (from first to upgrade -> last to upgrade) `Already offline` -> `No replicas` -> `# of leaders` -> `# of replicas` -> `not live` -> `live` -> `Overseer Node`.

There should likely be good documentation around this feature so that users are well aware of the expected behavior and how this mechanism works. 

The following options have been added to the SolrCloud CRD:

```golang
type SolrUpdateStrategy struct {
	// Method defines the way in which SolrClouds should be updated when the podSpec changes.
	// +optional
	Method SolrUpdateMethod `json:"method,omitempty"`

	// Options for Solr Operator Managed rolling updates.
	// +optional
	ManagedUpdateOptions ManagedUpdateOptions `json:"managed,omitempty"`
}

// SolrUpdateMethod is a string enumeration type that enumerates
// all possible ways that a SolrCloud can having rolling updates managed.
// +kubebuilder:validation:Enum=Managed;StatefulSet;Manual
type SolrUpdateMethod string

const (
	// Let the Solr Operator manage rolling updates to keep collections/shards available while updating pods in parallel.
	// This is the default option.
	ManagedUpdate SolrUpdateMethod = "Managed"

	// Use the default StatefulSet rolling updates logic. One pod at a time, starting with the highest ordinal.
	StatefulSetUpdate SolrUpdateMethod = "StatefulSet"

	// The Solr Operator and Kubernetes will not delete pods for updates. The user will be responsible for this.
	ManualUpdate SolrUpdateMethod = "Manual"
)

func (opts *SolrUpdateStrategy) withDefaults() (changed bool) {
	// You can't use an externalAddress for Solr Nodes if the Nodes are hidden externally
	if opts.Method == "" {
		changed = true
		opts.Method = ManagedUpdate
	}

	return changed
}

// Spec to control the desired behavior of managed rolling update.
type ManagedUpdateOptions struct {

	// The maximum number of pods that can be unavailable during the update.
	// Value can be an absolute number (ex: 5) or a percentage of the desired number of pods (ex: 10%).
	// Absolute number is calculated from percentage by rounding down.
	// If the provided number is 0 or negative, then all pods will be allowed to be updated in unison.
	//
	// Defaults to 25%.
	//
	// +optional
	MaxPodsUnavailable *intstr.IntOrString `json:"maxPodsUnavailable,omitempty"`

	// The maximum number of replicas for each shard that can be unavailable during the update.
	// Value can be an absolute number (ex: 5) or a percentage of replicas in a shard (ex: 25%).
	// Absolute number is calculated from percentage by rounding down.
	// If the provided number is 0 or negative, then all replicas will be allowed to be updated in unison.
	//
	// Defaults to 1.
	//
	// +optional
	MaxShardReplicasUnavailable *intstr.IntOrString `json:"maxShardReplicasUnavailable,omitempty"`
}
```

**Testing performed**

There are unit tests included that try to ensure that the pod selection and ordering logic is correct. As well as the parsing of the CRD options.

I have tested this manually, but haven't exhausted all possible cases. 
Before this is merged, this needs likely needs an integration test suite that ensures additional changes will not break this functionality.


docs/solr-cloud/managed-updates.md
```diff
@@ -0,0 +1,43 @@
+# Managed SolrCloud Rolling Updates
+
+Solr Clouds are complex distributed systems, and thus require a more delicate and informed approach to rolling updates.
+
+If the [`Managed` update strategy](solr-cloud-crd.md#update-strategy) is specified in the Solr Cloud CRD, then the Solr Operator will take control over deleting SolrCloud pods when they need to be updated.
+
+The operator will find all pods that have not been udpated yet and choose the next set of pods to delete for an udpate, given the following workflow.
```
thelabdude, line 7: s/udpated/updated ;-)

docs/solr-cloud/managed-updates.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Pod Update Workflow
+
+The logic goes as follows:
+
+1. Find the pods that are not up-to-date
+1. Retrieve the cluster state of the SolrCloud if there are any `ready` pods.
+    - If no pods are ready, then there is no endpoint to retrieve the cluster state from.
+1. Sort the pods in order of saftey for being restarted. [Sorting order reference](#pod-update-sorting-order)
```
thelabdude, line 16: s/saftey/safety

thelabdude:
LGTM +1 ... ran through various scenarios with a 5-node cluster and multiple collections. Also tried with no collection and a single node cluster just to verify no weird edges cases.

HoustonPutman:
Also tested some edge cases like using an image that doesn't exist, then reverting back. Works even better than the statefulset logic! 😄

sepulworld:
Need any help testing this out? Looks awesome. We would probably start using `ManagedUpdate` in production right from the get go! :)

HoustonPutman:
It'd be great if y'all could test it out. I think I want to add some stuff in the status about whether a node is up-to-date with the spec. (Right now it only looks at the image tag, not other pod specs). Other than that I think this is close to ready to merge.

Obviously every use case we have to test it out makes me even more confident! 😄

HoustonPutman:
@sepulworld The status changes are in. Let me know when you get a chance to test this out!

thelabdude (approved):
Great work as usual Houston! I kicked the tires on this PR in GKE over the past few days and looks great to me

bsankara (approved):

PR #144, opened 2020-09-30
Add options for ZK ACLs.

### LINKED ISSUE #143: Support Zookeeper ACLs
Currently it is somewhat cumbersome to add ZK ACLs for Solr. You can do it by specificying the correct Environement Variables in `SolrCloud.Spec.customSolrKubeOptions.podOptions.envVars`, but in order for the Solr process to recieve the envVars correctly they need to be included in SOLR_OPTS, which is placed in the wrong place to accept them in the pod envVar list.

It would also be nice for users to not have to know the nitty-gritty details on the exact env var names and config options. Having a simple section of the crd to specify a secret name, usernameKey and passwordKey to inject ZK ACLs would be optimal.

This is applicable for both the SolrPrometheusExporter and SolrCloud, but they do not necessarily have to be done concurrently.

HoustonPutman:
*Issue number of the reported bug or feature request: Resolves #143*

**Describe your changes**
Adding a section in the `SolrCloud.Spec.zkRef.connectionInfo` for specifying a secret that contains a username and password for ZK ACLs (both admin and read only).

```golang
// ZookeeperSpec defines the internal zookeeper ensemble to run for solr
type ZookeeperACL struct {
	// The name of the Kubernetes Secret that stores the username and password for the ACL.
	// This secret must be in the same namespace as the solrCloud is running in .
	SecretRef string `json:"secret"`

	// The name of the key in the given secret that contains the ACL username
	UsernameKey string `json:"usernameKey"`

	// The name of the key in the given secret that contains the ACL password
	PasswordKey string `json:"passwordKey"`
}
```

These ZK ACL options are also available for the Prometheus Exporter.

**Testing performed**
Unit tests have been added for injecting the ACLs. It has also been tested locally.


HoustonPutman:
Has unit tests, so the operator part is guaranteed to work. Also tested locally to make sure the solr and prometheus exporter params work correctly, as is expected.

bsankara (approved):

Issue #142, opened 2020-09-29
Cannot connect to ServiceIP service for either SolrCloud or SolrPrometheusExporter.


HoustonPutman:
**Describe the bug**
There is a bug with certain providers of Kube, where network requests fail if a Service's port name and targetPort name do not equal. 

More information can be found here: https://github.com/GoogleCloudPlatform/k8s-multicluster-ingress/issues/196

**Proposed Solution**
The solution is quite easy, change the servicePort to be the same name as the targetPort. For both SolrCloud and SolrPrometheusExporter Services.

This should not interrupt any traffic for solrClouds, as the Ingress rules use port numbers instead of port names.

PR #145, opened 2020-10-02
Use same portName as targetPort in service and fix Status bugs.

### LINKED ISSUE #141: The Cloud status has incorrect values for common and node addresses.
**Describe the bug**
The cloud status is being populated with incorrect information for node and common addresses.

The commonAddress has the wrong port number used, I believe it is using the nodePort instead of the commonPort.

The nodeAddress is using the node name instead of the pod name when constructing the first part of the URL.

HoustonPutman:
*Issue number of the reported bug or feature request: Fixes #141, #142*

**Describe your changes**
Two issues are fixed here:
1. The portName and targetPort for SolrCloud services have been changed to have the same name. This fixes issues with certain cloud providers that have bugs with regard to that. (#141)
1. The SolrCloud status object had multiple issues with the urls it displayed for users to connect with. (#142)

**Testing performed**
Unit tests have been modified to ensure that these values are now correct in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ Please visit the following pages for documentation on using and developing the S
 
 ## Version Compatibility & Upgrade Notes
 
+#### v0.2.7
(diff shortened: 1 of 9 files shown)


bsankara (approved):
LGTM. Thanks for this @HoustonPutman

PR #146, opened 2020-10-06
Integrate with cert-manager to issue TLS certs for Solr

thelabdude:
**Describe your changes**

_NOTE: I'll likely convert much of this info into user-facing docs in the .md files before merging_

Introduces an optional structure for the SolrCloud CRD to configure TLS for Solr pods. The `SolrTLSOptions` struct contains options to point the operator at secrets containing a PKCS12 keystore and keystore password. Optionally, the Solr operator can be configured to create a `Certificate` for Solr. Currently, the Solr operator only supports PKCS12 keystores; as of Java 8, the default key store format is PKCS12, which is an industry standard format for bundling an X.509 Certificate and private key and is considered an improvement over Java’s proprietary JKS format.

Behind the scenes, this feature relies on **cert-manager** (https://cert-manager.io/docs/), a Kubernetes controller for managing TLS certificates, including renewing certificates prior to expiration. As of cert-manager v0.15 (https://cert-manager.io/docs/release-notes/release-notes-0.15/#general-availability-of-jks-and-pkcs-12-keystores), cert-manager can create a PKCS12 (https://en.wikipedia.org/wiki/PKCS_12) keystore automatically when the `Certificate` is issued. Thus, we can simply have the cert-manager generate the PKCS12 keystore that the JVM can use directly without needing an `initContainer` to run a translation script.

The other benefit is that cert-manager supports pluggable Issuer implementations, including a self-signed Issuer for local development and a ACME compliant Issuer (https://tools.ietf.org/html/rfc8555) for working with services like **Let’s Encrypt** (https://letsencrypt.org/).

In general, there are **three basic use cases** the Solr operator needs to consider:

*1. Solr operator generated TLS Certificate*
User defines a cert-manager `ClusterIssuer` / `Issuer` and then Solr operator generates a certificate specifically for Solr using the specified Issuer:
```
 solrTLS:
    autoCreate:
      subjectDName: "O=MyOrg,OU=SomeUnit,C=USA"
      issuerRef:
        name: USER_DEFINED_ISSUER
        kind: Issuer | ClusterIssuer
```
_Note: don’t specify an *issuerRef* under *autoCreate* and the Solr operator will create a self-signed cert for dev purposes._

The benefit of letting the operator create the `Certificate` is it can configure the DNS names based on the operator's external addressability config or using the headless service DNS name if the SolrCloud is not exposed externally.

The Solr operator does not create other types of `Issuer` because requirements will differ based on the underlying provider. For instance, on GKE, to create a Let’s Encrypt Issuer you need a service account with various cloud DNS permissions granted for DNS01 challenges to work, see: https://cert-manager.io/docs/configuration/acme/dns01/google/. In general, we don’t want the Solr operator to assume responsibility for operations that should be done by the cert-manager. However, once the Issuer is configured, the Solr operator can automatically generate a Certificate for Solr.

*2. User generated TLS Certificate using cert-manager*
Users can also choose to just create the certificate using cert-manager directly; this gives users full control over the fields in the certificate. Once created, simply point the SolrCloud deployment at the TLS and keystore password secrets, e.g.
```
solrTLS:
    keyStorePasswordSecret:
      name: pkcs12-password-secret
      key: password-key
    pkcs12Secret:
      name: my-tls-cert
      key: keystore.p12
```
However, when generating the Certificate, users must take care to request that the pkcs12 keystore gets created using config similar to:
```
  keystores:
    pkcs12:
      create: true
      passwordSecretRef:
        key: password-key
        name: pkcs12-password-secret
```

*3. User supplied TLS Certificate*
Lastly, users may bring their own cert stored in a `kubernetes.io/tls` secret; for this use case, cert-manager is not required. There are many ways to get a certificate, such as from the GKE managed certificate process or from a CA directly. Regardless of how a Certificate is obtained, it needs to be stored in a Kubernetes TLS secret that contains a `.crt` file (x.509 certificate with a public key and info about the issuer) and a `.key` file (the private key).

A human operator can create the pkcs12 keystore needed by the Solr operator from the TLS secret using a process similar to the following: 
```
pass=$(kubectl get secret ${KEYSTORE_SECRET} -o jsonpath="{.data.password-key}" | base64 --decode)

kubectl get secret ${TLS_SECRET} -o jsonpath="{.data.tls\.crt}" | base64 --decode > tls.crt
kubectl get secret ${TLS_SECRET} -o jsonpath="{.data.tls\.key}" | base64 --decode > tls.key

openssl pkcs12 -export -in tls.crt -inkey tls.key -out keystore.p12 -passout "pass:${pass}"

kubectl create secret generic pkcs12-keystore-manual --from-file=./keystore.p12  --from-literal="password-key=${pass}"
```

Then simply configure the SolrCloud deployment to point to the secrets:
```
solrTLS:
    keyStorePasswordSecret:
      name: pkcs12-keystore-manual
      key: password-key
    pkcs12Secret:
      name: pkcs12-keystore-manual
      key: keystore.p12
```

*Ingress*
The Solr operator may create an nginx Ingress for exposing Solr pods externally. When TLS is enabled, the operator adds the following annotation and TLS settings to the Ingress manifest, such as:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/backend-protocol: HTTPS
spec:
  rules:
    ...
  tls:
  - secretName: tjp-selfsigned-cert-tls
```

*Certificate Renewal*

cert-manager automatically handles certificate renewal. From the docs:

> The default duration for all certificates is 90 days and the default renewal windows is 30 days. This means that certificates are considered valid for 3 months and renewal will be attempted within 1 month of expiration.
> 
>          https://docs.cert-manager.io/en/release-0.8/reference/certificates.html

However, this only covers updating the underlying TLS secret, which mounted secrets in the pod do get updated (see: https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically). However, the key and trust stores are read once during initialization of the JVM, so we need to recycle the Solr container in each pod to pick up the updated keystore.

The operator stashes the `resourceVersion` of the TLS secret in the `SOLR_TLS_SECRET_VERS` env var so that when the TLS secret changes, it will trigger an update on the StatefulSet. I'm guarding this behavior with an opt-in flag `RestartOnTLSSecretUpdate` as some users may not want to restart Solr pods when the TLS secret holding the cert changes and may instead choose to restart the pods during a maintenance window (presumably before the certs expire).

**Testing performed**
PR includes unit tests that verify correct configuration and reconcile behavior. The minimal config for self-signed certs for local development is shown below:
```
  solrTLS:
    autoCreate:
      subjectDName: "O=me"
```
For local dev, you’ll need to configure your browser to trust the certificate. For Mac, I used the process described here:
https://www.ateam-oracle.com/how-to-make-chrome-on-os-x-trust-a-self-signed-certificate

Or skip the browser and quickly test the cert using http:
```
http https://<ns>-<name>-solrcloud.ing.local.domain/solr/admin/info/system --verify=no
```
In addition, I tested integration with Let's Encrypt on GKE using the following config:

Create the ACME Issuer:
```
kind: Issuer
metadata:
  name: acme-letsencrypt-issuer
  namespace: search01
spec:
  acme:
    # The ACME server URL
    server: https://acme-staging-v02.api.letsencrypt.org/directory
    #server: https://acme-v02.api.letsencrypt.org/directory
    # Email address used for ACME registration
    email: SOMEBODY@example.com
    # Name of a secret used to store the ACME account private key
    privateKeySecretRef:
      name: acme-letsencrypt-issuer-pk
    solvers:
    - dns01:
        cloudDNS:
          project: solr-operator
          serviceAccountSecretRef:
            name: clouddns-dns01-solver-svc-acct
            key: key.json
```
This requires you to setup a service account with the appropriate cloudDNS permissions if you want to use DNS01 challenges, such as: `kubectl create secret generic clouddns-dns01-solver-svc-acct --from-file=key.json -n cert-manager`

Then, configure the Solr operator to use the `acme-letsencrypt-issuer`:
```
  solrTLS:
    autoCreate:
      name: search01-solr-tls
      subjectDName: "CN=search01.thelabdude.cloud, O=thelabdude, OU=k8s, C=USA"
      issuerRef:
        name: acme-letsencrypt-issuer
        kind: Issuer
```

**Additional context**
To install cert-manager:
```
helm repo add jetstack https://charts.jetstack.io
helm repo update

kubectl create ns cert-manager

helm upgrade --install cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --version v1.0.0 \
  --set installCRDs=true
```
You’ll need admin privileges to install the CRDs in a shared K8s cluster, so work with your friendly Kubernetes admin to install if needed (most likely cert-manager will already be installed).


thelabdude:
Thanks @HoustonPutman for figuring this out for me ;-)

thelabdude:
need to rebase the branch against master so just going to post a new clean PR

PR #162, opened 2020-12-06
Bring Go to 1.15

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>


HoustonPutman:
Nice, I like it. We recently upgraded to 1.14, but I think this will be a good idea after the next release.

PR #176, opened 2021-01-14
Standardize headers to be ASL2.0, move copyright to NOTICE

anshumg:


NOTICE.txt
```diff
@@ -0,0 +1,30 @@
+==============================================================
+ Apache Solr
```
HoustonPutman, line 2: Should this be Apache Solr Operator?

NOTICE.txt
```diff
@@ -0,0 +1,30 @@
...
+ Copyright 2006-2021 The Apache Software Foundation
+==============================================================
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+Includes software from other Apache Software Foundation projects,
+including, but not limited to:
+  - Apache Lucene Java
```
HoustonPutman, line 11: So this project doesn't actually contain software from Apache Lucene, right? We use the APIs, but not the libraries.

HoustonPutman: I imagine we are going to have to fill this out with the go dependencies we use?

HoustonPutman (commented):
Does `LICENSE` need to change too?
Or does that keep: `Copyright 2017 Bloomberg Finance L.P.`?

NOTICE.txt
anshumg: Yes, but we might not have to add Solr as it's the same project. I think the Lucene bit that exists in 'Solr' is just cruft from 10 years ago :) It'll be required again in the future, but for the operator, we shouldn't need Lucene in this list.

anshumg:
@HoustonPutman  - reading more about the license and the requirements around that. We might be able to remove that too. I'll do that as another PR if needed once I have clarity.

The only thing missing here is the fix in `check-license.sh`

HoustonPutman (approved):
Good to go.

We can fill in the NOTICE later when we figure out the dependency licensing bti.

anshumg:
FYI, still waiting for Infra to remove the required checks before I can squash and merge.

klaporte:
Looks good to me. Thanks @anshumg  & @HoustonPutman

klaporte (approved):

anshumg:
Thanks for taking a look @klaporte :)

Issue #173, opened 2021-01-14
Helm repository is not available anymore


dalbani:
We used to use the Helm repository located at https://bloomberg.github.io/solr-operator/charts, but it seems not to be available anymore.
Is there a new location? The documentation still mentions the old location.

benediktarnold:
I don't know when it happened, but since solr-operator is now an apache project, it makes sense that the chart isn't hosted on bloombergs github page anymore. I'm looking for an alternative as well.

vladiceanu:
@bsankara any news on this? thanks

HoustonPutman:
Hey everyone, sorry about this. I'll try to have it available as soon as possible.

Unfortunately the repo was moved to `apache/lucene-solr-operator` instead of `apache/solr-operator`. So it will have to be moved again. But I will try to make it work in the meantime. I'll post an update here and on the slack channel when it should be available.

HoustonPutman:
In the meantime, you should be able to use this instead: (Just a fix until the real changes have been made)

```
helm repo add solr-operator https://raw.githubusercontent.com/apache/lucene-solr-operator/master/docs/charts
```

dalbani:
But, as you said, will the Git project and Helm repo eventually move to `apache/solr-operator` instead of `apache/lucene-solr-operator`?

HoustonPutman:
Great question, I should have addressed it.

It looks like we will unfortunately not be able to move to apache/solr-operator for a while. It will eventually move there, but I cant give a timeline. We will definitely be more vocal about the change beforehand next time and try to make the transition smoothless.

dalbani:
Okay, great, thanks a lot.

anshumg:
Just to be clear, @HoustonPutman meant 'smooth' 😄

janhoy:
Did you ask INFRA if they can rename the repo to solr-operator for us? They will understand why...

HoustonPutman:
Yeah, I did. They do permissioning based off of apache project name prefix. So since Solr isn't a TLP yet, there are no user groups called "solr", so no one would have permissions for the "solr-operator". Since the group is called "lucene", we are able to have permissions for "lucene-solr-operator". We will be able to migrate once the "solr" TLP user groups are created, irrespective of when the Solr code actually splits.

It is frustrating, but it should be an easy enough move whenever the TLP gets created. (hopefully sooner rather than later)

PR #178, opened 2021-01-14
Change the repo location to apache

HoustonPutman:
Fixes: #173


anshumg (approved):
LGTM.
There are a bunch of other things that could've been removed but makes sense to just limit this change to the repo location.

Issue #183, opened 2021-01-15
Migrate repo from Bloomberg to Apache


HoustonPutman:
This includes:

- Changing the CRD groups from solr.bloomberg.com to solr.apache.org
- Updating documentation
- Changing the docker image location
- Updating code paths and go module

HoustonPutman:
We need to ensure that documentation is clear for new users trying to install the v0.2.8 of the bloomberg Solr Operator, since v0.3.0 (apache) is not yet available.

The documentation, as well as the examples directory, currently includes examples that use the `solr.bloomberg.com` API group CRDs, and we want to make sure the documentation continue to works for users until the new version of the operator is released. This will likely require creating "released" documentation that is independent of the current latest version of the repo.

PR #186, opened 2021-01-15
Cleanup README.md

anshumg:
Part of: #183 

Fix Bloomberg references.

Things pending:
- Docker Pulls - need to wait until we push an image to docker hub
- Slack - should we cross ref the Lucene/Solr one too?
- Contributing Guidelines - should we link this here? https://cwiki.apache.org/confluence/display/solr/HowToContribute ? This isn’t a 100% correct considering building and running are completely different for the operator. We need a new doc for this perhaps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Commit since last release](https://img.shields.io/github/commits-since/apache/lucene-solr-operator/latest.svg)](https://github.com/apache/lucene-solr-operator/commits/master)
 [![Docker Pulls](https://img.shields.io/docker/pulls/bloomberg/solr-operator)](https://hub.docker.com/r/bloomberg/solr-operator/)
 [![Slack](https://img.shields.io/badge/slack-join_chat-white.svg?logo=slack&style=social)](https://kubernetes.slack.com/messages/solr-operator)
+[![Mailing List]]
 
 The __Solr Operator__ manages Apache Solr Clouds within Kubernetes. It is built on top of the [Kube Builder](https://github.com/kubernetes-sigs/kubebuilder) framework.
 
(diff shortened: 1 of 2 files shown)


HoustonPutman:
I agree we should have a separate HowToContribute for the operator. I have some basic dev documentation, but it's not complete.

HoustonPutman:
And agreed on the docker images, that should be a separate PR entirely.

README.md
```diff
@@ -120,23 +121,15 @@ Have you had a good experience with the **Solr Operator**? Why not share some lo
...
 We welcome issue reports [here](../../issues); be sure to choose the proper issue template for your issue, so that we can be sure you're providing the necessary information.
 
-Before sending a [Pull Request](../../pulls), please make sure you read our
-[Contribution Guidelines](https://github.com/bloomberg/.github/blob/master/CONTRIBUTING.md).
-
 ## License
 
 Please read the [LICENSE](LICENSE) file here.
 
-## Code of Conduct
```
HoustonPutman, line 130: Is there no Apache code of conduct we can use here instead?

HoustonPutman (commented):
Is the Solr slack workspace (the non ASF one) actually a thing yet? We can add it, but I think the Kube slack channel is fine for now. I don't have a huge opinion on it though

README.md
klaporte: What about this? https://www.apache.org/foundation/policies/conduct

anshumg: I was trying to look up something that I thought we used for PR creation in the `lucene-solr` project, just to reuse and be consistent but if I'm unable to find it, this will work :)

anshumg: @HoustonPutman - we shouldn't add any non-ASF communication channels here as they aren't official.

PR #189, opened 2021-01-19
Override Prometheus exporter config XML via a user-supplied ConfigMap

thelabdude:
Allow users to override the Prometheus exporter config (XML) via a user-provided ConfigMap supplied via the existing `prometheusExporter.Spec.CustomKubeOptions.ConfigMapOptions.ProvidedConfigMap` setting.

Keep the md5 hash of the custom config XML in an annotation on the pod spec of the deployment so that a rolling restart is triggered when the configuration changes.

### DIFF
--- a/controllers/solrprometheusexporter_controller.go
+++ b/controllers/solrprometheusexporter_controller.go
@@ -117,6 +117,11 @@ func (r *SolrPrometheusExporterReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 			return ctrl.Result{}, err
 		}
 
+		// capture the MD5 for the default config XML, otherwise we already computed it above
+		if configXmlMd5 == "" {
+			configXmlMd5 = fmt.Sprintf("%x", md5.Sum([]byte(configMap.Data[configMapKey])))
+		}
+
 		// Check if the ConfigMap already exists
 		configMapLogger := logger.WithValues("configMap", configMap.Name)
 		foundConfigMap := &corev1.ConfigMap{}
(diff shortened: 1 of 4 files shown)


controllers/solrprometheusexporter_controller.go
```diff
@@ -92,10 +124,18 @@ func (r *SolrPrometheusExporterReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 		if err != nil && errors.IsNotFound(err) {
 			configMapLogger.Info("Creating ConfigMap")
 			err = r.Create(context.TODO(), configMap)
+
+			// capture the MD5 for the config XML
+			configXml, _ := configMap.Data[configMapKey]
+			configXmlMd5 = fmt.Sprintf("%x", md5.Sum([]byte(configXml)))
```
HoustonPutman, line 130: You need to move this before `if/else`, otherwise the xml annotation will be removed when the reconcile is called, but the configMap is not updated or created.

See: https://github.com/apache/lucene-solr-operator/pull/188/files#diff-27cb619f80d28ac158f7885b7925d8582babfa7fe1f824599d65a745302d237cR222

controllers/solrprometheusexporter_controller_test.go
```diff
@@ -200,6 +200,7 @@ func TestMetricsReconcileWithExporterConfig(t *testing.T) {
 	testMapsEqual(t, "deployment labels", util.MergeLabelsOrAnnotations(expectedDeploymentLabels, testDeploymentLabels), deployment.Labels)
 	testMapsEqual(t, "deployment annotations", testDeploymentAnnotations, deployment.Annotations)
 	testMapsEqual(t, "pod labels", util.MergeLabelsOrAnnotations(expectedDeploymentLabels, testPodLabels), deployment.Spec.Template.ObjectMeta.Labels)
+	testPodAnnotations["solr.apache.org/exporterConfigXmlMd5"] = "2c354e0aeb176a6b5560f46fe35c9ca7"
```
HoustonPutman, line 203: Could you possibly do this dynamically, like: https://github.com/apache/lucene-solr-operator/pull/188/files#diff-c7a739f50a172286ab35c4729f5e1795249b1be34bd386695c9facb05d437fb4R274

I moved the `expectConfigMap` to be above the statefulSet check, so that I could use the data that was already fetched to make the Md5.

controllers/solrprometheusexporter_controller.go
thelabdude: ah yes ... great catch, and thanks for fixing for solr.xml

thelabdude:
Still need an explicit unit test for the user-provided config before this one is ready to merge.

HoustonPutman (approved):
This should be good to go!

PR #191, opened 2021-01-20
Make docs stable based on the last version, store in Github Pages.

### LINKED ISSUE #187: Host documentation that represents the most recent release
The documentation is currently hosted in the default branch, so that it gets updated along-side new features. 

The problem with this however is that the documentation does not reflect the most recently released version, but instead the latest unstable changes in the repo. As #183 states, this is especially a problem now that the latest documentation is going to be incompatible with the latest released version.

Documentation needs to be versioned, and likely needs to be hosted outside of the github repository. However the first step of action is to have a source of documentation that is only updated on new version releases.

HoustonPutman:
Fixes #187 

This points documentation to the Github Pages version, which will only be updated on version upgrades.


HoustonPutman:
You can also check [the github pages branch](https://github.com/apache/lucene-solr-operator/tree/gh-pages) to make sure everything is good there.

PR #193, opened 2021-01-21
Support custom log4j2 config from user-provided ConfigMap

thelabdude:
Fixes: https://github.com/apache/lucene-solr-operator/issues/192

SolrCloud controller now looks for the presence of a `log4j2.xml` key in the `instance.Spec.CustomSolrKubeOptions.ConfigMapOptions.ProvidedConfigMap`

The only tricky aspect of this PR is we don't save the md5 in an annotation if the provided Log4j2 config includes `monitorInterval=` as we don't need to trigger a restart of the Solr pods as Log4j2 will reload the config at runtime. If the config does not include `monitorInterval`, then we save the md5 hash of the XML in an annotation on the pod spec to trigger a restart as we do for solr.xml

Also, cleaned up some constant string usage around `solr.xml` and `log4j2.xml`


controllers/solrcloud_controller.go
```diff
@@ -182,44 +182,61 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
 		} else {
-			return requeueOrNot, fmt.Errorf("Provided ConfigMap %s has no data",
-				instance.Spec.CustomSolrKubeOptions.ConfigMapOptions.ProvidedConfigMap)
+			return requeueOrNot, fmt.Errorf("Provided ConfigMap %s has no data", providedConfigMapName)
 		}
-	} else {
+	}
+
+	if configMapInfo[util.SolrXmlFile] == "" {
```
HoustonPutman, line 231: So if a user passes a custom configMap that just contains a log4j.xml, the operator will still create a configMap for the xml, correct?

controllers/util/solr_util.go
```diff
@@ -327,13 +330,42 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
...
+			}
+			podAnnotations[LogXmlMd5Annotation] = configMapInfo[LogXmlMd5Annotation]
+		}
+
+		// cannot use /var/solr as a mountPath, so mount the custom log config in a sub-dir
+		volName := "log4j2-xml"
+		mountPath := fmt.Sprintf("/var/solr/%s-log-config", solrCloud.Name)
+		log4jPropsEnvVarPath := fmt.Sprintf("%s/%s", mountPath, LogXmlFile)
+
+		solrVolumes = append(solrVolumes, corev1.Volume{
```
HoustonPutman, line 348: So if the custom configMap contains both a Solr.xml and a log4j.xml, then this will add 2 volumes, using the same configMap, correct?

Is this supported in Kube? I think there might be an issue. But if you've tested it, then maybe it's ok to go.

It would be pretty easy to make sure that the volume is only added once, and it has the correct `items`.

thelabdude: This is a good catch! K8s allows it and results in a structure like the following in the STS:
```
      volumes:
      - configMap:
          defaultMode: 420
          items:
          - key: solr.xml
            path: solr.xml
          name: dev-custom-solr-xml
        name: solr-xml
      - configMap:
          defaultMode: 420
          items:
          - key: log4j2.xml
            path: log4j2.xml
          name: dev-custom-solr-xml
        name: log4j2-xml
```
But we certainly should use a single volume with multiple `items` as you suggest ;-) Will fix it up and add a test for both keys being provided.

controllers/solrcloud_controller.go
thelabdude: If the user-provided ConfigMap only contains a key for the log4j2.xml, then the operator will create the default ConfigMap for solr.xml, resulting in this structure in the STS for a SolrCloud named `dev`:
```
      volumes:
      - configMap:
          defaultMode: 420
          items:
          - key: solr.xml
            path: solr.xml
          name: dev-solrcloud-configmap
        name: solr-xml
```
The `if configMapInfo[util.SolrXmlFile] == "" {` is just how we know there wasn't a `solr.xml` in the user-provided ConfigMap at that point. Does that sound right?

HoustonPutman: Yeah, that sounds great. Just wanted to make sure I understood correctly.

thelabdude: ok cool ... I switched to keeping all the user-provided ConfigMap info in a single map arg to `GenerateStatefulSet` vs. having multiple function args just to keep things tidy.

controllers/util/solr_util.go
thelabdude: Latest commit addresses this issue.

controllers/suite_test.go
```diff
@@ -76,6 +78,9 @@ func SetupTestReconcile(inner reconcile.Reconciler) (reconcile.Reconciler, chan
 	requests := make(chan reconcile.Request)
 	fn := reconcile.Func(func(req reconcile.Request) (reconcile.Result, error) {
 		result, err := inner.Reconcile(req)
+		if err != nil {
+			fmt.Printf("\n\nReconcile Error: %s\n\n", err)
```
thelabdude, line 82: this is a bit ugly but so useful in tracking down errors during reconcile that are returned like this:
```
				if !strings.Contains(solrXml, "${hostPort:") {
					return requeueOrNot,
						fmt.Errorf("Custom solr.xml in ConfigMap %s must contain a placeholder for the 'hostPort' variable, such as <int name=\"hostPort\">${hostPort:80}</int>",
							providedConfigMapName)
				}
```
W/o this change, the error isn't reported and the test just fails with reconcile timeouts. I think it's a useful addition.

thelabdude:
Now with docs! ;-)

controllers/solrcloud_controller_test.go
```diff
@@ -759,18 +754,246 @@ func TestCloudWithCustomSolrXmlConfigMapReconcile(t *testing.T) {
...
+
+	// Create the user-provided ConfigMap first to streamline reconcile,
+	// other tests cover creating it after the reconcile loop starts on the SolrCloud
+	userProvidedConfigMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testCustomConfigMap,
+			Namespace: instance.Namespace,
+		},
+		Data: map[string]string{
+			util.LogXmlFile:  "<Configuration/>",
```
HoustonPutman, line 939: Should there be a test where the log4J file is auto-updated? (Thus no need for an annotation)

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -127,4 +127,118 @@ the solr-operator can manage Zookeeper ensemble(s) for you.
...
+```
+kubectl annotate pod -l technology=solr-cloud --list=true
+```
+If the custom `solr.xml` changes in the user-provided ConfigMap, then the operator triggers a rolling restart of Solr pods to apply the updated configuration settings automatically.
+
+To summarize, if you need to customize `solr.xml`, provide your own version in a ConfigMap and changes made to the XML in the ConfigMap are automatically applied to your Solr pods.
+
+### Custom Log Configuration
+
+As with `solr.xml`, the operator configures Solr with a default Log4j2 configuration file: `-Dlog4j.configurationFile=/var/solr/log4j2.xml`.
```
HoustonPutman, line 215: Unlike with `solr.xml`, the operator uses the default log4j2 configuration file that ships with Solr.

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -127,4 +127,118 @@ the solr-operator can manage Zookeeper ensemble(s) for you.
...
+apiVersion: v1
+metadata:
+  name: custom-solr-xml
+data:
+  solr.xml: |
+    <?xml version="1.0" encoding="UTF-8" ?>
+    <solr>
+      ... CUSTOM CONFIG HERE ...
+    </solr>
+```
```
HoustonPutman, line 185: I would put a bolded statement here stating the need for a `<hostPort>${hostPort}</hostPort>` under the <solr><solrcloud> section.

HoustonPutman: Also say that your statefulset and solr pods will not be created if this is missing from the XML.

controllers/solrcloud_controller.go
```diff
@@ -182,44 +182,61 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-		// ConfigMap doesn't have to have a solr.xml, but if it does, then it needs to be valid!
 		if foundConfigMap.Data != nil {
-			solrXml, ok := foundConfigMap.Data["solr.xml"]
-			if ok {
+			logXml, hasLogXml := foundConfigMap.Data[util.LogXmlFile]
+			solrXml, hasSolrXml := foundConfigMap.Data[util.SolrXmlFile]
+
+			// if there's a user-provided config, it must have one of the expected keys
+			if !hasLogXml && !hasSolrXml {
+				return requeueOrNot, fmt.Errorf("User provided ConfigMap %s must have one of 'solr.xml' and/or 'log4j2.xml'",
```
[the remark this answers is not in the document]
thelabdude: Can you point me to an example elsewhere in the code for what you're looking for here?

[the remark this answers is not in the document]
HoustonPutman: https://github.com/apache/lucene-solr-operator/blob/main/controllers/solrcloud_controller.go#L330

Just a simple thing for us to look for when implementing events.

HoustonPutman (approved):
Looks good to me!

PR #195, opened 2021-01-22
Improve Prom exporter docs

thelabdude:
Add instructions for using the Prometheus exporter with Prometheus stack

### DIFF
--- a/docs/solr-prometheus-exporter/README.md
+++ b/docs/solr-prometheus-exporter/README.md
@@ -50,3 +50,187 @@ The Prometheus Exporter can be setup to scrape a standalone Solr instance.
 In order to use this functionality, use the following spec field:
 
 `SolrPrometheusExporter.spec.solrRef.standalone.address`
+
+## Prometheus Stack
+
+In this section, we'll walk through how to use the Prometheus exporter with the [Prometheus Stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack).
+
+The Prometheus Stack provides all the services you need for monitoring Kubernetes applications like Solr and is the recommended way of deploying Prometheus and Grafana.
+
+### Install Prometheus Stack
+
+Begin by installing the Prometheus Stack in the `monitoring` namespace with Helm release name `mon`:
+```
+MONITOR_NS=monitoring
+PROM_OPER_REL=mon
+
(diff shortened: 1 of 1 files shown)


docs/solr-prometheus-exporter/README.md
```diff
@@ -49,4 +49,188 @@ All ACL fields are **required** if an ACL is used.
...
+```
+kubectl get svc dev-prom-exporter-solr-metrics --output jsonpath="{@.spec.ports}"
+```
+You should see output similar to:
+```
+[{"name":"solr-metrics","port":80,"protocol":"TCP","targetPort":8080}]
+```
+
+### Create a Service Monitor
+The Prometheus operator (deployed with the Prometheus stack) uses service monitors to find which services to scrape metrics from. Thus, we need to define a service monitor for our exporter service `dev-prom-exporter-solr-metrics`.
```
HoustonPutman, line 189: I would add a comment that this differs from the Kubernetes discovery configuration provided by base prometheus: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#service

HoustonPutman (approved):
This is awesome stuff!!

It would be great in the future when we have a prometheus-exporter helm chart, to be able to auto-install the serviceMonitor for users as well (if desired).

docs/solr-prometheus-exporter/README.md
```diff
@@ -49,4 +49,188 @@ All ACL fields are **required** if an ACL is used.
...
+GRAFANA_POD_ID=$(kubectl get pod -l app.kubernetes.io/name=grafana --no-headers -o custom-columns=":metadata.name" -n monitoring)
+kubectl port-forward -n monitoring $GRAFANA_POD_ID 3000
+```
+Open Grafana using `localhost:3000` and login with username `admin` and password `prom-operator`.
+
+Once logged into Grafana, import the Solr dashboard JSON corresponding to the version of Solr you're running.
+
+Solr does not export any useful metrics until you have at least one collection defined.
+
+_Note: Solr 8.8.0 includes an updated dashboard that includes better metrics for query performance._
```
anshumg, line 236: May be switch this to `Solr 8.8.0 and newer versions`

anshumg (approved):
This looks great! Thanks @thelabdude .

Issue #194, opened 2021-01-21
Change the Solr Operator docker namespace to "Apache"


HoustonPutman:
A part of #183

The Docker image currently resides in the `bloomberg/` namespace in DockerHub. We need to replace this with `apache/`.

The current documentation and helm chart should not need any changes, because `v0.2.8` is still hosted under `bloomberg/` and that will not change. This change will only be reflected with `v0.3.0` and after.

Even though the name of the project is currently `apache/lucene-solr-operator`, we want to make sure that the docker image is named `apache/solr-operator` for future-compatibility.

PR #199, opened 2021-01-22
Change docker image location to apache/

HoustonPutman:
Fixes: #194 

The non-deployed helm chart will no longer work with the default image name, but that's fine because the released helm chart will continue to work.


anshumg (approved):
👍🏽 

I've added this: https://github.com/apache/lucene-solr-operator/pull/201

PR #201, opened 2021-01-22
Fix docker image location in README.md

anshumg:
A part of: #194


HoustonPutman:
So I think we should only change this one when we actually deploy it, because that link will just be broken otherwise.

anshumg:
Yup, just created this for the future,

anshumg:
Actually, let me mark this as draft so we don't accidentally merge this.

HoustonPutman:
We have to say goodbye to our 200k download count 😢 

But hopefully we make it back soon!

PR #207, opened 2021-01-29
docs: removing step in local tutorial for collection creation

sepulworld:


HoustonPutman:
Instead, I changed the instructions to use the Collections API to create the collection. That way users have a better understanding of how to interact with the cloud 😃

PR #209, opened 2021-02-03
Basic TLS Support

### LINKED ISSUE #150: Add configuration options to enable TLS on Solr pods using cert-manager issued certificates
Should be possible to enable TLS for encrypted transport between Solr pods.

Leverage cert-manager (https://cert-manager.io/docs/) to obtain a PKCS12 formatted keystore in a TLS secret and then configure the SOLR_SSL_* env vars to load the cert into each Solr pod.

thelabdude:
Fixes #150 

Taking a simplified approach than what was attempted in #151 by removing `autoCreate` mode and instead relying on users to create the TLS secret however they see fit. The approach in this PR avoids introducing a dependency on cert-manager, although we still recommend users use cert-manager if they don't already have a TLS certificate solution in place.


api/v1beta1/solrcloud_types.go
```diff
@@ -971,3 +996,47 @@ type SolrCloudList struct {
...
+type SolrTLSOptions struct {
+	// TLS Secret containing a pkcs12 keystore
+	PKCS12Secret *corev1.SecretKeySelector `json:"pkcs12Secret"`
+
+	// Secret containing the key store password; this field is required as most JVMs do not support pkcs12 keystores without a password
+	KeyStorePasswordSecret *corev1.SecretKeySelector `json:"keyStorePasswordSecret"`
+
+	// Determines the client authentication method, either None, Want, or Need;
+	// this affects K8s ability to call liveness / readiness probes so use cautiously.
+	// +optional
```
HoustonPutman, line 1018: You can use `// +kubebuilder:default=None`, and get rid of the `withDefaults()` method.

api/v1beta1/solrcloud_types.go
```diff
@@ -52,6 +52,10 @@ const (
 
 	SolrTechnologyLabel      = "solr-cloud"
 	ZookeeperTechnologyLabel = "zookeeper"
+
+	DefaultKeyStorePath         = "/var/solr/tls"
```
HoustonPutman, line 56: These can go in solr_util.go right? I don't think they are used in the API.

controllers/solrcloud_controller.go
```diff
@@ -261,12 +263,65 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			}
+
+			if _, ok := foundTLSSecret.Data[instance.Spec.SolrTLS.PKCS12Secret.Key]; !ok {
+				// the keystore.p12 key is not in the TLS secret, indicating we need to create it using an initContainer
+				needsPkcs12InitContainer = true
+			}
+		}
+
+		// create a job to set the cluster property but we need a ZK connection first
+		urlSchemeJob := util.GenerateUrlSchemeJob(instance, &newStatus)
```
HoustonPutman, line 304: Does this have to happen before the Solr nodes are started?

controllers/solrcloud_controller.go
```diff
@@ -381,7 +436,7 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	if !reflect.DeepEqual(instance.Status, newStatus) {
 		instance.Status = newStatus
-		logger.Info("Updating SolrCloud Status")
+		r.Log.Info("Updating SolrCloud Status: ", "namespace", instance.Namespace, "name", instance.Name, "status", instance.Status)
```
HoustonPutman, line 439: can you keep the logger.Info() and just add the "status" part? It'll keep the logging consistent.

controllers/solrcloud_controller.go
```diff
@@ -716,7 +768,9 @@ func (r *SolrCloudReconciler) SetupWithManagerAndReconciler(mgr ctrl.Manager, re
 		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.StatefulSet{}).
 		Owns(&corev1.Service{}).
-		Owns(&extv1.Ingress{})
+		Owns(&extv1.Ingress{}).
+		Owns(&corev1.Secret{}).
```
HoustonPutman, line 772: We just watch secrets, not create them I think. This will require that lovely `indexAndWatchForTlsSecrets()` method. I can handle this though!

controllers/solrcloud_controller.go
```diff
@@ -76,6 +77,8 @@ func UseZkCRD(useCRD bool) {
 // +kubebuilder:rbac:groups=solr.apache.org,resources=solrclouds/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=zookeeper.pravega.io,resources=zookeeperclusters,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=zookeeper.pravega.io,resources=zookeeperclusters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
HoustonPutman, line 80: For secrets, we should be able to get away with: `verbs=get;list;watch`. I think...

HoustonPutman (approved):
This is good to go! Awesome work, as usual.

PR #216, opened 2021-02-15
NT: Allow the definition of multiple imagePullSecrets.

jward-bw:
Currently you can only specify a single imagePullSecret in the ContainerImage, for the SolrImage in SolrCloudSpec. This allows a user to define any number of imagePullSecrets, which is required if initContainers or sidecarContainers need different imagePullSecrets to the SolrImage.

### DIFF
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -19,15 +19,16 @@ package util
 
 import (
 	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
 	solr "github.com/apache/lucene-solr-operator/api/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"sort"
-	"strconv"
-	"strings"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


jward-bw:
Sorry for the multiple failed runs, didn't read the docs enough on building the manifests.

jward-bw:
Was I supposed to add `helm/solr-operator/crds/crds.yaml` to my commit? A diff that size smells suspiscious.

controllers/util/solr_util.go
```diff
@@ -520,8 +520,15 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 	}
 
 	if solrCloud.Spec.SolrImage.ImagePullSecret != "" {
-		stateful.Spec.Template.Spec.ImagePullSecrets = []corev1.LocalObjectReference{
-			{Name: solrCloud.Spec.SolrImage.ImagePullSecret},
+		if nil != customPodOptions {
+			stateful.Spec.Template.Spec.ImagePullSecrets = append(
+				customPodOptions.ImagePullSecrets,
```
HoustonPutman, line 525: I think customPodOptions secrets will only be added if the `SolrImage.ImagePullSecret` is provided. You should probably have it added in a separate if statement.

HoustonPutman (changes requested):
Great start! 

Would you mind adding a test for this? And also implement the functionality in the prometheusExporter code as well?

The helm chart change is definitely odd. Are you using the kustomize version specified in `hack/install_dependencies.sh`?

jward-bw:
> Are you using the kustomize version specified in hack/install_dependencies.sh?

Yep that was the issue, I had an older version installed.

> ... implement the functionality in the prometheusExporter code as well?

Done!

> Would you mind adding a test for this?

Sorry this is my first time writing go code, and I'm not familiar with the testing structure. Where should tests for these changes go?

controllers/util/solr_util.go
jward-bw: Fixed! Thanks for pointing that out

HoustonPutman (commented):
Ok the code looks great!

As for the tests, here is some help:
- [Where we have been specifying customPodOptions test values](https://github.com/apache/lucene-solr-operator/blob/main/controllers/controller_utils_test.go#L535).  
So you should be able to add a list of ImagePullSecrets here to use other places in the tests.
- [An example test that you could add in the imagePullSecrets](https://github.com/apache/lucene-solr-operator/blob/main/controllers/solrcloud_controller_test.go#L188)
- [Where you would actually test that they got added to the Pod](https://github.com/apache/lucene-solr-operator/blob/main/controllers/solrcloud_controller_test.go#L280)

You can do the same thing for the prometheus exporter tests:
- [An example prom-exp test that you could add in the imagePullSecrets](https://github.com/apache/lucene-solr-operator/blob/main/controllers/solrprometheusexporter_controller_test.go#L58)
- [Where you would actually test that they got added to the prom-exp Pod](https://github.com/apache/lucene-solr-operator/blob/main/controllers/solrprometheusexporter_controller_test.go#L126)

These tests aren't actually making sure that it runs correctly, it makes sure that the resulting statefulSet object that is created looks as it should. So you just need to make sure that the pod Template in the statefulset has the correct imagePullSecrets section.

jward-bw:
> As for the tests, here is some help:


Thanks so much for the pointers Houston, let me know if I've missed anything.

HoustonPutman (approved):
Don't think you've missed anything! This looks great, feature & tests. 👌 

Thanks for the contribution! Hopefully the first of many 😉

PR #221, opened 2021-02-17
Work with basic-auth enabled SolrCloud clusters

### LINKED ISSUE #218: Enhancements for working with basic auth enabled Solr clusters
For background on Solr basic auth, see: 
https://lucene.apache.org/solr/guide/8_8/basic-authentication-plugin.html#basic-authentication-plugin

The Solr operator should provide an easy way to enable basic authentication for a SolrCloud cluster.

There are two main uses cases we should support:

1. The operator should bootstrap a default `security.json` with basic auth enabled. This allows users to easily spin-up a secure Solr cluster without having to know all the details of how Solr security works. Having the Solr API be protected by basic-auth out of the box leads to a better getting started experience for Solr and the operator in general.
2. The operator should allow users to supply their own `security.json` (via ConfigMap) and corresponding secret with credentials the operator should use to make API requests to Solr. Both are needed and if the user chooses to supply their own `security.json` then they need to make sure the credentials are in the secret as well. For this use case, the operator's main job is to bootstrap the `security.json` in ZK as part of the initialization process.

When basic auth is enabled, we need to ensure the liveness / readiness probes work (`/admin/info/system` endpoint). In addition, the rolling restart logic requires access to the `/admin/collections` endpoint. Also, backup / restore requires API access. Thus, the Solr operator needs access to a username and password to make API calls. These credentials will come from a Kubernetes secret in the same namespace.

For the first use case (bootstrap a default config), the only tricky part is we need to create the user password hashes in the same manner that Solr expects them. The default `security.json` will define an `admin` user, a general purpose read-only `solr` user, and special `k8s-oper` user that *only* has read access to the API endpoints needed to operator in Kubernetes, such as the `/admin/info/system` endpoint used by probes. The password for each user can be retrieved using `kubectl` using:
```
kubectl get secret dev-solrcloud-basic-auth -o jsonpath='{.data.k8s-oper}' | base64 --decode
```

Also, I'm not sure if the operator should watch the `security.json` ConfigMap for updates (as we do with other user-provided ConfigMaps) and try to apply the updated JSON? There's an API for updating the security config and the ref guide suggests using that so maybe the operator just bootstraps the file and then it's up to users to maintain via the API? Trying to accommodate user updates via the API and/or ConfigMap could get messy.

For the probes, K8s allows specifying a header for HTTP endpoints, but then our password will live in the STS config. I think it would be better to set an env var from the secret and then run a command for the probe using this env var. The SolrCLI command-line tool provides an `api` action that we can use for the probes:
```
java -Dbasicauth=k8s-oper:****** \
  -Dsolr.httpclient.builder.factory=org.apache.solr.client.solrj.impl.PreemptiveBasicAuthClientBuilderFactory \
  -Dsolr.install.dir="/opt/solr" \
  -Dlog4j.configurationFile="/opt/solr/server/resources/log4j2-console.xml" \
  -classpath "/opt/solr/server/solr-webapp/webapp/WEB-INF/lib/*:/opt/solr/server/lib/ext/*:/opt/solr/server/lib/*" \
  org.apache.solr.util.SolrCLI api -get http://localhost:8983/solr/admin/info/system
```
Unfortunately, `bin/solr` does not expose the `api` action :-( So we'll need to take some liberties on how the probe command is constructed, such as assuming Solr is installed at `/opt/solr`

Still testing / tweaking the default `security.json` but I imagine it would be something like this:
```
{
      "authentication":{
        "blockUnknown": true,
        "class":"solr.BasicAuthPlugin",
        "credentials": {
          "admin":  "...",
          "k8s-oper":     "...",
          "solr":   "..."
        },
        "realm":"Solr Basic Auth",
        "forwardCredentials": false
      },
      "authorization": {
         "class": "solr.RuleBasedAuthorizationPlugin",
         "user-role": {
            "admin": ["admin", "probe"],
            "k8s-oper": ["probe"],
            "solr": ["users", "probe"]
         },
         "permissions": [
           { "name": "k8s-probes", "role":"probe", "collection": null, "path":"/admin/info/system" },
           { "name": "k8s-status", "role":"probe", "collection": null, "path":"/admin/collections" },
           { "name": "all", "role":["admin","users"] },
           { "name": "read", "role":["admin","users"] },
           { "name": "update", "role":["admin"] },
           { "name": "security-read", "role": "admin"},
           { "name": "security-edit", "role": "admin"}
         ]
      }
    }
```

thelabdude:
This PR fixes #218 

The changes in this PR allow users to bootstrap a `security.json` for a SolrCloud cluster with basic auth enabled.
```
spec:
  ...
  solrSecurity:
    authenticationType: Basic
```
A word about the probes ... we recommend configuring Solr to allow un-authenticated access over HTTP to the probe endpoint(s) and the bootstrapped `security.json` does this for you automatically. However, if you want secure the probe endpoints, then you need to set `probesRequireAuth: true`:
```
spec:
  ...
  solrSecurity:
    authenticationType: Basic
    probesRequireAuth: true
```
When auth is enabled, the liveness & readiness probes are configured to execute a command instead of using HTTP. Of course you can supply the credentials in an `Authorization` header for the HTTP probes, but that would require a restart of all pods if the password changes. With a command, we can load the credentials from file that comes from a secret. K8s will update the secret file when the secret changes automatically. You can see this work by changing the password in Solr first (using the security API), e.g.
```
curl --user $CREDS http://localhost:8983/solr/admin/authentication \
   -H 'Content-type:application/json' -d '{"set-user": {"k8s-oper":"Test1234"}}'
```
If you wait a bit, then probes will start to fail b/c the password has changed on the Solr side but is not yet reflected in the operator / k8s layer. Edit the secret to update the password for the `k8s-oper` user and you'll see probes begin to work again after a brief delay. If you look behind the scenes, you'll notice the command used for probes invokes the `SolrCLI` utility:
```
java  org.apache.solr.util.SolrCLI api -get http://localhost:8983/solr/admin/info/system
```
Unfortunately, the `bin/solr` command does not expose the `api` action we need, so we have to invoke `SolrCLI` directly. 

See user docs included in this PR for additional info ...

**TODO:**
- [x] User docs
- [x] Auth for Backup / Restore API
- [x] Prometheus Exporter config
- [x] Mount secret used by probe command as file (instead of env var) so that K8s automatically updates it when the secret changes


controllers/util/solr_util.go
```diff
@@ -971,10 +991,26 @@ func generateZKInteractionInitContainer(solrCloud *solr.SolrCloud, solrCloudStat
...
+
+	volumeMounts := []corev1.VolumeMount{}
+	if configMapInfo[SecurityJsonFile] != "" {
+		// mount security.json from the ConfigMap at /tmp/security.json
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{Name: "security-json", MountPath: "/tmp"})
+		if cmd == "" {
+			cmd += "solr zk ls ${ZK_CHROOT} -z ${ZK_SERVER} || solr zk mkroot ${ZK_CHROOT} -z ${ZK_SERVER}; "
+		}
+		cmd += "ZK_SECURITY_JSON=$(/opt/solr/server/scripts/cloud-scripts/zkcli.sh -zkhost ${ZK_HOST} -cmd get /security.json); "
+		cmd += "if [ ${#ZK_SECURITY_JSON} -lt 3 ]; then /opt/solr/server/scripts/cloud-scripts/zkcli.sh -zkhost ${ZK_HOST} -cmd putfile /security.json /tmp/security.json; echo \"put security.json in ZK\"; fi"
```
thelabdude, line 1010: haha ... don't laugh at this ;-) it was the easiest way to check if the znode data is empty "" or {} using /bin/sh

controllers/util/solr_util.go
```diff
@@ -1158,3 +1195,170 @@ func createZkConnectionEnvVars(solrCloud *solr.SolrCloud, solrCloudStatus *solr.
...
+	pass[len(pass)-1] = upper[rand.Intn(len(upper))]
+	perm := rand.Perm(len(chars))
+	for i := 1; i < len(pass)-1; i++ {
+		pass[i] = chars[perm[i]]
+	}
+	return pass
+}
+
+// this mimics the password hash generation approach used by Solr
+func solrPasswordHash(passBytes []byte) string {
```
thelabdude, line 1351: See Solr's Sha256AuthenticationProvider: https://github.com/apache/lucene-solr/blob/master/solr/core/src/java/org/apache/solr/security/Sha256AuthenticationProvider.java#L114

HoustonPutman:
A few thoughts:

- I completely agree that only files used in the filesystem in Solr (solr.xml, log4j, prometheus exporter config) should be stored in configMaps. Those that are merely meant to bootstrap Zookeeper, should not be stored in configMaps.
- I like the initAnonymousEndpoints, but just so that I'm clear, that's just for the bootstrapped security.json right? Even if the user changes that list after the cluster is created, it won't be reflected in the security.json? If so maybe we leave it up to the users to use the security APIs to manage that. Not really sure yet either way.
- For the HTTP Probes, would it not be easy to merely remove the auth from those initially? Or I guess we can't tell whether the user has removed auth for it or not, so we need to provide it by default.

controllers/solrcloud_controller.go
```diff
@@ -256,6 +257,55 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			}
+			if err != nil {
+				return requeueOrNot, err
+			}
+			// supply the bootstrap security.json to the initContainer via a simple BASE64 encoding env var
+			configMapInfo[util.SecurityJsonFile] = string(basicAuthSecret.Data[util.SecurityJsonFile])
+		}
+
+		// save some basic information needed to configure probes for auth-enabled Solr pods
+		configMapInfo[util.SecuritySecret] = basicAuthSecret.Name
```
HoustonPutman, line 301: shouldn't you only set this if probes should use the auth data?

Also maybe we should rename this from configMapInfo to something else...

controllers/solrprometheusexporter_controller.go
```diff
@@ -286,6 +301,13 @@ func (r *SolrPrometheusExporterReconciler) SetupWithManagerAndReconciler(mgr ctr
 		return err
 	}
 
+	// Get notified when the basic auth secret updates; exporter pods must be restarted if the basic auth password
+	// changes b/c the credentials are loaded from a Java system property at startup and not watched for changes.
+	ctrlBuilder, err = r.indexAndWatchForBasicAuthSecret(mgr, ctrlBuilder)
```
HoustonPutman, line 306: Honestly think at this point we should bake something into ControllerRuntime to automate this. We use it so often.

(Nothing to do with this PR, just an observation)

controllers/solrprometheusexporter_controller.go
```diff
@@ -209,7 +209,22 @@ func (r *SolrPrometheusExporterReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
-	deploy := util.GenerateSolrPrometheusExporterDeployment(prometheusExporter, solrConnectionInfo, configXmlMd5, tlsClientOptions)
+	basicAuthMd5 := ""
+	if prometheusExporter.Spec.SolrReference.BasicAuthSecret != nil {
+		selector := prometheusExporter.Spec.SolrReference.BasicAuthSecret
+		basicAuthSecret := &corev1.Secret{}
+		err := r.Get(context.TODO(), types.NamespacedName{Name: selector.Name, Namespace: prometheusExporter.Namespace}, basicAuthSecret)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		passBytes, ok := basicAuthSecret.Data[selector.Key]
```
HoustonPutman, line 220: We might have to look in two places here. It could be in `data` or `stringData`.

https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret

api/v1beta1/solrcloud_types.go
```diff
@@ -1022,3 +1044,35 @@ type SolrTLSOptions struct {
...
+	// are secured.
+	//
+	// If you don't supply this secret, then the operator bootstraps a default security.json file and creates a
+	// corresponding secret containing the credentials for three users: admin, solr, and k8s-oper. All API requests
+	// from the operator are made as the 'k8s-oper' user, which is configured with minimal access. The 'solr' user has
+	// basic read access to Solr resources. Once the security.json is bootstrapped, the operator will not update it!
+	// You're expected to use the 'admin' user to access the Security API to make further changes. It's strictly a
+	// bootstrapping operation.
+	// +optional
+	BasicAuthSecret *corev1.SecretKeySelector `json:"basicAuthSecret,omitempty"`
```
HoustonPutman, line 1063: I am not familiar with how other CRDs handle this, but should the "key" have to be the username, or should it be passed in separately?

Like `basicAuthUsername` and `basicAuthPasswordSecret`. 

[Looking at the kubernetes docs](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret), it looks like the standard for Basic Auth secrets is to store "username": <> and "password": <>.

api/v1beta1/solrprometheusexporter_types.go
```diff
@@ -88,6 +89,13 @@ type SolrReference struct {
 	// Settings to configure the SolrJ client used to request metrics from TLS enabled Solr pods
 	// +optional
 	SolrTLS *SolrTLSOptions `json:"solrTLS,omitempty"`
+
+	// If Solr is secured, you'll need to provide credentials for the Prometheus exporter to authenticate.
+	// The key is the username. If basic auth is enabled on the SolrCloud instance, the default secret (unless you are
+	// supplying your own) is named using the pattern: <SOLR_CLOUD_NAME>-solrcloud-basic-auth. If using the security.json
+	// bootstrapped by the Solr operator, then the username should be "k8s-oper".
+	// +optional
+	BasicAuthSecret *corev1.SecretKeySelector `json:"basicAuthSecret,omitempty"`
```
HoustonPutman, line 98: Same comment about basicAuth secrets standards.

controllers/solrcloud_controller.go
```diff
@@ -256,6 +257,55 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		secretName := instance.BasicAuthSecretName()
+		basicAuthSecret := &corev1.Secret{}
+
+		// user has the option of providing a secret with credentials the operator should use to make requests to Solr
+		if sec.BasicAuthSecret != nil {
+			if err := r.Get(ctx, types.NamespacedName{Name: secretName, Namespace: instance.Namespace}, basicAuthSecret); err != nil {
+				return requeueOrNot, err
+			}
+
+			if _, ok := basicAuthSecret.Data[sec.BasicAuthSecret.Key]; !ok {
```
HoustonPutman, line 274: Could be in Data or StringData

controllers/util/solr_util.go
```diff
@@ -1158,3 +1195,200 @@ func createZkConnectionEnvVars(solrCloud *solr.SolrCloud, solrCloudStatus *solr.
...
+	var annotations map[string]string
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        solrCloud.BasicAuthSecretName(),
+			Namespace:   solrCloud.GetNamespace(),
+			Labels:      labels,
+			Annotations: annotations,
+		},
+		Data: generateSecurityJson(solrCloud),
+		Type: "Opaque",
```
HoustonPutman, line 1245: We might want to use `Type: "kubernetes.io/basic-auth"`?

controllers/solrcloud_controller.go
```diff
@@ -256,6 +257,55 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			// supply the bootstrap security.json to the initContainer via a simple BASE64 encoding env var
+			configMapInfo[util.SecurityJsonFile] = string(basicAuthSecret.Data[util.SecurityJsonFile])
+		}
+
+		// save some basic information needed to configure probes for auth-enabled Solr pods
+		configMapInfo[util.SecuritySecret] = basicAuthSecret.Name
+		configMapInfo[util.SecurityUsername] = username
+
+		// need the creds below for getting CLUSTERSTATUS
+		creds := username + ":" + string(basicAuthSecret.Data[username])
```
HoustonPutman, line 305: does the Data need to be decoded from base64?

HoustonPutman (commented):
Reviewed everything except for solr_util.go.

Main question I have for now is the format of basic auth secrets that we should accept. It seems like [Kube has a standard](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret).

api/v1beta1/solrcloud_types.go
thelabdude: We're storing creds for multiple users: admin, solr, and k8s-oper as well as the security.json (so we don't have to recompute it for every reconcile loop). So our secret is not a K8s basic auth secret

api/v1beta1/solrcloud_types.go
```diff
@@ -1022,3 +1044,35 @@ type SolrTLSOptions struct {
...
+	// +optional
+	ProbesRequireAuth bool `json:"probesRequireAuth,omitempty"`
+
+	// A list of endpoints that allow un-authenticated (aka "anonymous") access; this allows you to open Solr for
+	// un-authenticated access to query endpoints but lock down all other requests. This setting only applies during
+	// initial bootstrapping of the security.json file. Changing this after security.json has been applied by the
+	// operator to a SolrCloud instance has no effect. Obviously, if you're supplying your own basicAuthSecret, then this
+	// setting does not apply as you're expected to configure your own security.json.
+	// +optional
+	InitAnonymousEndpoints []string `json:"initAnonymousEndpoints,omitempty"`
```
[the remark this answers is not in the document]
thelabdude: ok, I removed it

controllers/solrprometheusexporter_controller.go
thelabdude: afaik, `stringData` is a convenience for YAML and secrets always store all data in `data` but I'll verify if they supply via `stringData` that we can read it from `data`

HoustonPutman: Ahhh very cool, did not know that!

thelabdude: Confirmed that only `data` is stored ... here's an example:
```
apiVersion: v1
stringData:
  k8s-oper: Test1234
kind: Secret
metadata:
  name: user-provided-basic-auth
type: Opaque
```
and `kubectl get secret user-provided-basic-auth -o yaml`  gives:
```
apiVersion: v1
data:
  k8s-oper: VGVzdDEyMzQ=
kind: Secret
metadata:
  name: user-provided-basic-auth
type: Opaque
```
and in the logs, we have:
```
2021-02-19T19:38:56.486Z	INFO	controllers.SolrCloud	For user-supplied secret, password is:	{"namespace": "dev", "solrCloud": "dev", "password": "Test1234"}
```
_Note: I just added that log message to test this scenario ... it has since been removed, i.e. we're not logging passwords_

api/v1beta1/solrcloud_types.go
thelabdude: One idea I liking more and more is to have 2 secrets, one that is a `"kubernetes.io/basic-auth"` that holds the creds for the `k8s-oper` user and another `Opaque` that holds the bootstrapped `admin` user and the `security.json`. We can name the latter `<cloud>-solrcloud-security-bootstrap` (or similar) which makes it clear it's only for bootstrapping security and the former holds the creds for the user needed by the operator.

sylus:
Hi there I ended up testing this P.R. and it works great in a non Istio enabled namespace.

However the problem is with the solr-zk-setup init container when is in an istio enabled namespace as there is no network connectivity.

The Istio CNI plugin may cause networking connectivity problems for any application initContainers. When using Istio CNI kubelet starts an injected pod with the following steps:

1. The Istio CNI plugin sets up traffic redirection to the Istio sidecar proxy within the pod
2. All init containers execute and complete successfully
3. The Istio sidecar proxy starts in the pod along with the pods other containers

Init containers execute before the sidecar proxy starts, which can results in traffic loss during their exection. Some workarounds include:

a) Set the traffic.sidecar.istio.io/excludeOutboundIPRanges
b) Set the traffic.sidecar.istio.io/excludeOutboundPorts
c) Running initContainers as user 1337 since that bypasses Istio

I am waiting for zookeeper-operator to allow for custom annotations to be passed then this should work.

thelabdude:
@sylus thanks for testing, really appreciate it ... I'll try to work in a solution using your work-arounds

sylus:
I also had do this:

https://github.com/apache/lucene-solr-operator/issues/223

To get the quorum property but Solr then worked talking to ZooKeeper when was added (except for init container)

And thanks so much for the awesome work it was pretty cool to see the basic auth setup just work and not have to do anything :)

controllers/solrcloud_controller.go
thelabdude: I renamed it to `reconcileConfigInfo` since it holds additional config info computed during reconcile ... ;-)

thelabdude:
@sylus Let's tackle the Istio related changes in a follow-on PR from this one, i.e. no work-arounds for that here ;-) Can help with the follow-on PR once this is merged though, should be close ...

sylus:
Thanks @thelabdude that makes complete sense to me :D

sylus:
Once pravega allows for custom annotations technically wouldn't need any changes to your P.R. so that might be the best route forward :D

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -543,3 +543,234 @@ spec:
...
+```
+kubectl get secret <CLOUD>-solrcloud-security-bootstrap -o jsonpath='{.data.solr}' | base64 --decode
+```
+
+The operator makes requests to secured Solr endpoints as the `k8s-oper` user; credentials for the `k8s-oper` user are stored in a separate secret of type `kubernetes.io/basic-auth`
+with name `<CLOUD>-solrcloud-basic-auth`. The `k8s-oper` user is configured with read-only access to a minimal set of endpoints, see details in the **Authorization** sub-section below.
+Remember, if you change the `k8s-oper` password using the Solr security API, then you **must** update the secret with the new password or the operator will be locked out.
+Also, changing the password for the `k8s-oper` user in the K8s secret after bootstrapping will not update Solr! You're responsible for changing the password in both places.
+
+#### Liveness and Readiness Probes
```
HoustonPutman, line 596: I think this would fit better as a `###` section after `Option 2: ...`

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -543,3 +543,234 @@ spec:
...
+
+### Prometheus Exporter with Basic Auth
+
+If you enable basic auth for your SolrCloud cluster, then you need to point the Prometheus exporter at the basic auth secret containing the credentials for making API requests to `/admin/metrics` and `/admin/ping` for all collections.
+
+```
+spec:
+  solrReference:
+    basicAuthSecret: user-provided-secret
+```
```
HoustonPutman, line 773: I would re-iterate here the endpoints needed for the prometheus exporter auth user, when moving it to the prometheus exporter docs.

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -543,3 +543,234 @@ spec:
...
+}
+```
+A few aspects of the default `security.json` configuration warrant a closer look. First, the `probesRequireAuth` setting 
+(defaults to `false`) governs the value for `blockUnknown` (under `authentication`) and whether the probe endpoint(s) require authentication:
+```
+      {
+        "name": "k8s-probe-0",
+        "role": null,
+        "collection": null,
+        "path": "/admin/info/system"
```
HoustonPutman, line 693: We probably want to add `/admin/info/health` for the future as well. (this would affect multiple places in the PR)

controllers/util/solr_util.go
```diff
@@ -1157,3 +1193,240 @@ func createZkConnectionEnvVars(solrCloud *solr.SolrCloud, solrCloudStatus *solr.
...
+	return "Basic " + b64.StdEncoding.EncodeToString([]byte(creds))
+}
+
+func ValidateBasicAuthSecret(basicAuthSecret *corev1.Secret) error {
+	if basicAuthSecret.Type != corev1.SecretTypeBasicAuth {
+		return fmt.Errorf("invalid secret type %v; user-provided secret %s must be of type: %v",
+			basicAuthSecret.Type, basicAuthSecret.Name, corev1.SecretTypeBasicAuth)
+	}
+
+	if _, ok := basicAuthSecret.Data[corev1.BasicAuthUsernameKey]; !ok {
```
HoustonPutman, line 1243: I think these two are validated by the API server, but not a bad thing to have it here as well.

HoustonPutman (commented):
Looks really good! Just a few small things to finish this up

controllers/util/solr_util.go
thelabdude: Sadly not the case ... I tried creating a basic auth secret w/o the username and K8s happily obliged me :(

controllers/solrcloud_controller.go
```diff
@@ -256,6 +259,74 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+					reconcileConfigInfo[util.SecurityJsonFile] = string(bootstrapSecret.Data[util.SecurityJsonFile])
+				}
+
+				basicAuthSecret = authSecret
+			}
+			if err != nil {
+				return requeueOrNot, err
+			}
+
+			if reconcileConfigInfo[util.SecurityJsonFile] == "" {
```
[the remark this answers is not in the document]
thelabdude: ok good idea, let me test that ... I think it would still trigger a full rolling restart of the STS (b/c the initContainer def changes) but you're right we should fail if they want to delete it

api/v1beta1/solrcloud_types.go
```diff
@@ -1022,3 +1040,37 @@ type SolrTLSOptions struct {
...
+type AuthenticationType string
+
+const (
+	Basic AuthenticationType = "Basic"
+)
+
+type SolrSecurityOptions struct {
+	// Indicates the authentication plugin type that is being used by Solr; for now only "Basic" is supported by the
+	// Solr operator but support for other authentication plugins may be added in the future.
+	AuthenticationType AuthenticationType `json:"authenticationType,omitempty"`
```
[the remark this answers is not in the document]
thelabdude: I considered that and chose to make it required by the user to set to `Basic` but I don't have strong feelings about enforcing that, will make optional with default `Basic`

docs/solr-cloud/solr-cloud-crd.md
thelabdude: We do you think we should do about the `customPodOptions` probes and auth? Currently, if the user supplies a custom  probe config with a different path and has `probesRequireAuth`, the code is not picking up the custom path. If `probesRequireAuth` is `false`, then the custom path does get added to the authz rules correctly. This is a bit trappy but I don't know if the operator should try to swap out their custom probe config with a command either? So perhaps we just doc this limitation for now?

controllers/util/solr_util.go
HoustonPutman: Very strange. Well good to have the check then!

controllers/solrcloud_controller.go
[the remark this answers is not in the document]
HoustonPutman: definitely would cause a full rolling restart, but at least their cloud would be able to reconcile 😅

api/v1beta1/solrcloud_types.go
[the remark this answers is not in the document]
HoustonPutman: Don't have strong feelings either. Up to you!

docs/solr-cloud/solr-cloud-crd.md
thelabdude: If we go with the punt on `probesRequireAuth` and custom probe paths approach, I think we should also error out during reconcile so it's clear we don't like their config, i.e. something along the lines of:
```
Found custom probe HTTP paths and probesRequireAuth=true, this configuration is not currently supported. Please switch to probesRequireAuth=false when using custom probe HTTP paths.
```

thelabdude: I've included `/admin/info/health` as a default in the rules for now

thelabdude: Re-worked this a bit and improved the docs in this area. Added a check for how custom probe paths are handled to one of the existing unit tests

HoustonPutman (approved):
This is good to go. Thoroughly tested it locally, and couldn't find any issues.

Awesome feature, and huge win for the community!

Thanks for the great work @thelabdude!!

PR #224, opened 2021-02-22
Various fixes - NPE, logging

HoustonPutman:
- Fix logging for Prometheus exporter
- Fix Null Pointer Exception when using additional volumes
- Add generated file missing from previous commit


thelabdude (approved):
Looks good!

PR #225, opened 2021-02-24
Fix Ingress API group for RBAC rules.

jward-bw:
Helm chart no longer works as of this PR:
https://github.com/apache/lucene-solr-operator/pull/217

This should fix the issue.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -143,7 +143,7 @@ rules:
   - patch
   - update
 - apiGroups:
-  - extensions
+  - networking.k8s.io
   resources:
   - ingresses
   verbs:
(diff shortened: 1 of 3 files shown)


HoustonPutman (approved):
Great find!

That file is actually auto-generated via the comment tags in solrcloud-controller, and the command `make manifests`.

I've made the change and run the command.

jward-bw:
> Great find!
> 
> That file is actually auto-generated via the comment tags in solrcloud-controller, and the command `make manifests`.
> 
> I've made the change and run the command.

Aha cheers Houston. We had just pulled the previous change in and found the error. Is `make manifests` something you would usually expect to be done as part of a PR like yours, or something that we should be running before deploying?

HoustonPutman:
`make manifests` should always be done before submitting a PR. It's a part of the PR check, but I guess there are some file paths we don't check for the manifests changes...

You should be able to use the `main` branch to deploy by default without running another command. If you are changing code locally, then definitely run manifests before deploying for tests.

PR #226, opened 2021-02-25
Add terminationGracePeriod option, and use it when killing Solr.

### LINKED ISSUE #227: Errors often occur while deleting Solr Pods
Its very common for Kubernetes to spew the following error when trying to take down a Solr pod:

> Exec lifecycle hook ([solr stop -p 8983]) for Container "solrcloud-node" in Pod "..." failed - error: command 'solr stop -p 8983' exited with 137: , message: "Sending stop command to Solr running on port 8983 ... waiting up to 180 seconds to allow Jetty process 12 to stop gracefully."

This is likely because Solr and Kubernetes have a different idea of the amount of time to wait before trying to forcefully kill a pod. Solr defaults to a time of 180 seconds, as shown in the message above. The Solr Operator defaults to giving the Solr statefulset a `terminationGracePeriodSeconds` of 10. Therefore Kubernetes will try to kill the pod much earlier than Solr thinks it has time to wait.

We should make sure that the `terminationGracePeriodSeconds` of the statefulSet and the `SOLR_STOP_WAIT` are always in sync with each other (not necessarily the exact same value), and that the default is more reasonable than 10 seconds.

HoustonPutman:
Fixes #227 

This will use the `terminationGracePeriod`, minus a few seconds, and give that to Solr as the `SOLR_STOP_WAIT` value.

That way Solr and Kubernetes are on the same page as to how much time to wait before forcefully stopping Solr.

`terminationGracePeriod` used to be statically set to 10 seconds. Now it defaults to 30 seconds, but is configurable.


controllers/util/solr_util.go
```diff
@@ -266,6 +270,9 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 	solrHostName := solrCloud.AdvertisedNodeHost("$(POD_HOSTNAME)")
 	solrAdressingPort := solrCloud.NodePort()
 
+	// Solr can take longer than SOLR_STOP_WAIT to run solr stop, give it a few extra seconds before forcefully killing the pod.
+	solrStopWait := terminationGracePeriod - 5
```
madrob, line 274: Check for positive value here, somebody might have set grace period on their cluster to 5s or less.

HoustonPutman: maybe we set a minimum possible value for terminationGracePeriod? Like 10 seconds?

HoustonPutman: Added the minimum as well as a check for a negative value, in case we remove the minimum in the future.

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -810,3 +812,22 @@ Also, changing the password for this user in the K8s secret will not update Solr
...
+
+The default time given is 60 seconds, before Solr or Kubernetes tries to forcefully stop the Solr process.
+You can override this default with the field:
+
+```yaml
+spec:
+  ...
+  customSolrKubeOptions:
+    podOptions:
+      terminationGracePeriodSeconds: 120
```
madrob, line 832: So this still means that administrators need to set this manually, rather than automatically picking up what is derived from the k8s source of truth?

HoustonPutman: This is a setting that users can set on pods, we merely forward this information to the pod. There is no k8s source of truth other than the value that each pod has. The default, set by the Solr Operator is 60 seconds.

Also I think the default value (for the pod) is different based on the kubernetes distribution you use, but we shouldn't be relying on a default that has nothing to do with Solr.

madrob: If a user issues a command `k delete --grace-period=5s` than that will override our configured setting. We should figure out how to do our best to react to the new directives.

HoustonPutman: That command is kind of similar to kill -9. I'm not sure there is a way to protect against that. Is there a way to figure out what grace period is given when your pod is trying to be deleted?

Most of the time, Pods are going to be deleted by Kubernetes (the statefulset controller) or the solrcloud controller, during managed upgrades. Neither of those use a different grace period, so the usage here is safe.

madrob:
Please add a `since 0.3.0` somewhere in the docs for this.

PR #229, opened 2021-03-01
Resolve dependency license handling

### LINKED ISSUE #228: Handle dependency licenses correctly
Before we release v0.3.0 under apache, we need to make sure that our release correctly deals with dependency licenses.

HoustonPutman:
Resolves #228 

The intent of this PR is to resolve the dependency license issue, so that docker images built for the Solr Operator, as well as source releases, follow ASF policy.

Changes:
- Docker image shouldn't need a source release, as well as a release with all of the dependency code vendored. We instead will package all golang dependency licenses (and code for licenses that require that source code be provided with any released binaries), along with the solr-operator binary.
- Some necessary additions to the NOTICES file

Other fixes include linting for the licenses, and having a pre-PR make target: `make prepare`.


HoustonPutman:
This approach sounds good to go according to the feedback in [LEGAL-562](https://issues.apache.org/jira/browse/LEGAL-562)

dependency_licenses.csv
```diff
@@ -0,0 +1,53 @@
+cloud.google.com/go/compute/metadata,Unknown,Apache-2.0
```
anshumg, line 1: Do we need to store this file (in this format) ?

.github/workflows/docker.yaml
```diff
@@ -17,4 +17,4 @@ jobs:
 
       # Cleanup & Install dependencies
       - run: docker --version
-      - run: make docker-vendor-build
\ No newline at end of file
+      - run: make docker-build
```
anshumg, line 20: nit: new line

anshumg (commented):
Thanks for taking care of this, Houston. This looks great.

Just left a few minor comments. I'm not sure if we need to store the licenses csv file.

dependency_licenses.csv
HoustonPutman: If not, that's fine. We could remove the whole workflow around this. If you don't think it's necessary, then I can tear it all out.

I think it's good to have so that we can keep track of the dependency Licenses that PRs are adding.

anshumg: Makes sense :)

Let's keep it here.

anshumg (approved):

Issue #222, opened 2021-02-19
Add docs for migrating from bloomberg (v0.2.8) to apache (v0.3.0)


HoustonPutman:
Migrating from v0.2.8 to v0.3.0 will require manual intervention, because the resource groups are changing.

Documentation is needed to make this transition as smooth as possible for users.

PR #230, opened 2021-03-02
Update the owner reference of owned resources, if the reference is not correct

HoustonPutman:
Relates to #222 

This change is necessary if in-place updates from Bloomberg resources to Apache resources are to succeed.

With this change, resources created for Bloomberg SolrClouds can be claimed and re-controlled by Apache SolrClouds.

This does make new resource management code more complex, but it is a good feature to have, probably.


thelabdude (approved):
Really cool solution here!

PR #231, opened 2021-03-02
Add conditional dependency for zk-operator helm chart

chaicesan:
This PR is the first of two phases that will allow us to make use of up-to-date versions of the zk-operator as well as being able to install it using a conditional helm chart dependency.

The aim of this PR is just to add the ability to install the latest version of the zk-operator via a helm chard dependency. Later work will involve making use of the new latest zk crd.

There were some pain points when it came to updating the dependencies, mainly due to compatibility involving `k8s.io/client-go`. I'm still pretty new to go and go modules so let me know if I am doing anything suspect. 

This is my first PR to this project so please let me know if I have missed anything. 

Just to test I installed the operator with helm using a kubernetes cluster on a dev environment. I also built and deployed the image (with the changes in this pr) to a private repo made reference to using a custom `values.yaml` file.
```
solr-operator-zookeeper-operator-5864b989d6-96m9z   1/1     Running     0          2m37s
solr-operator-6596fbb445-97tmm                      1/1     Running     0          2m37s
solr-solrcloud-zookeeper-0                          1/1     Running     0          54s
solr-solrcloud-zookeeper-1                          1/1     Running     0          42s
solr-solrcloud-zookeeper-2                          1/1     Running     0          23s
solr-solrcloud-1                                    1/1     Running     0          54s
solr-solrcloud-2                                    1/1     Running     0          54s
solr-solrcloud-0                                    1/1     Running     0          54s
```


helm/solr-operator/Chart.yaml
```diff
@@ -95,4 +95,10 @@ annotations:
...
         image:
-          tag: 8.7.0
\ No newline at end of file
+          tag: 8.7.0
+
+dependencies:
+  - name: 'zookeeper-operator'
+    version: 0.2.9
+    repository: https://charts.pravega.io
+    condition: useZkOperator
```
HoustonPutman, line 104: We probably want to change this to a different variable, such as `zookeeper-operator.enabled`, because we want users to be able to use the zookeeper-operator, without installing it as a dependency. That requires two variables. Then when setting the command line variable for solr, we can do an OR of `zookeeper-operator.enabled` and `useZkOperator`.

I'm flexible on names, if you don't think that's clear.

HoustonPutman (changes requested):

helm/solr-operator/Chart.yaml
chaicesan: Tell me what you think of the names I have chosen.

HoustonPutman: [The helm docs](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) give some good guidance. I think this logic is very good in general. We should just use the name `zookeeper-operator` instead of `zookeeperOperator`, since that's the name of the dependency.

We should also think about maybe including backwards compatibility of the old `useZookeeperOperator` option, which was unfortunately a string `"true"`.

I can take a stab at updating the documentation around this.

HoustonPutman:
This is starting to look very good! I'll give it a test locally soon, to make sure that there are no issues with the Zookeeper Cluster that is created in the new version of the zookeeper-operator.

HoustonPutman:
Hey @chaicesan , I went ahead and changed the variable name myself, and added backwards compatibility with the old `useZkOperator` option. I also updated various docs across the project.

I'm not so sure about using Chart.lock and keeping the `charts/zookeeper-operator-0.2.9.tgz` and `Chart.lock` in the repo. That's something that can be generated at release time, when building the chart. Is there a reason why you included it in the PR specifically?

chaicesan:
@HoustonPutman Sorry for the delay. I agree that it can be done on release so I will remove them today.

HoustonPutman:
No worries at all @chaicesan , I've mainly just been taking care of random documentation and licensing issues.

I think this is almost ready to go. I'm going to do some testing first to make sure that we don't need to change the Solr operator code at all, and everything runs smoothly. I have a bet we will need to do at least one thing.

HoustonPutman:
I think we are very close to having this merge ready.

I updated the zk-connection string to use the new URLs used by this version of the ZK Operator. We just need to make all the other fields update cleanly. We should probably auto-populate the kubeDomain field from the SolrCloud Spec into the ZK Cluster spec as well... But that's very very easy.

HoustonPutman:
I think I'm good now. Will let it sit for a day or two, and then merge! Reach out if you think there is anything missing.

Once again, thanks for all of your great work @chaicesan!!

chaicesan:
Nice one! Thanks for your help on this too @HoustonPutman :)

HoustonPutman (approved):

PR #232, opened 2021-03-03
Add note about versions supporting log4j.xml

madrob:
Would be good to add this to more of the features, but I don't know have enough context to label everything correctly.


madrob:
Thanks for the additional updates @HoustonPutman. It looks good to me, anything else we would want before committing?

HoustonPutman:
I think it's good to go!

PR #241, opened 2021-03-23
Help for upgrading from v0.2.x to v0.3.0

### LINKED ISSUE #222: Add docs for migrating from bloomberg (v0.2.8) to apache (v0.3.0)

HoustonPutman:
Resolves #222 

This includes a test that installs v0.2.6, upgrades to v0.2.8 and then upgrades to v0.3.0.

The documentation gives two options, upgrading in an environment managed manually, and an environment managed via a system like Ansible.

### DIFF
--- a/hack/test-upgrade.sh
+++ b/hack/test-upgrade.sh
@@ -4,14 +4,14 @@
 kubectl delete crds --all; kubectl delete pvc --all; helm delete solr-operator apache
 helm ls -a --all-namespaces | awk 'NR > 1 { print  "-n "$2, $1}' | xargs -L1 helm delete
 
-helm repo add apache-solr https://apache.github.io/lucene-solr-operator/charts && helm repo update
+helm repo add apache-solr https://solr.apache.org/charts && helm repo update
 
 ####
 ##  Setup an environment using the v0.2.6 Solr Operator
 ####
 
 # Install the Zookeeper Operator
-kubectl apply -f https://apache.github.io/lucene-solr-operator/example/dependencies/zk_operator.yaml
+kubectl apply -f https://apache.github.io/solr-operator/example/dependencies/zk_operator.yaml
 
 # Install a past version of the Solr Operator
 helm install solr-operator apache-solr/solr-operator --version 0.2.6 --set ingressBaseDomain=localhost.com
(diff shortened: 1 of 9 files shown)


hack/test-upgrade.sh
```diff
@@ -0,0 +1,172 @@
...
+####
+##  Install the Apache Solr Operator separately
+####
+# Build the Apache release (TODO: Change to release when it is cut)
+make docker-build
+helm dependency build helm/solr-operator
+
+# Uninstall Old ZK Operator
+echo "Deleting all Zookeeper Operator resources, except for CRDs" && \
+  kubectl delete deployment zk-operator && \
```
thelabdude, line 123: for me the deployment is named `zookeeper-operator` vs. `zk-operator`

HoustonPutman: How were you installing it for Solr Operator v0.2.x?

thelabdude: I just did it independently of the solr-operator:
```
helm upgrade --install zookeeper-operator pravega/zookeeper-operator \
  --version=0.2.9 --namespace=${NAMESPACE}
```

HoustonPutman: Ahhh if you are running it 0.2.9 already, you should just use `--set zookeeper-operator.install=false --set zookeeper-operator.use=true`

Probably good to add to the docs.

thelabdude:
Should this PR change the chart version in `helm/solr-operator/Chart.yaml` to `0.3.0`? Cool if not ... just confused me when I installed the chart from a checkout of this PR and got:
```
helm ls
NAME              	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART                   	APP VERSION
apache            	tjp      	1       	2021-03-24 11:07:21.514488 -0600 MDT	deployed	solr-operator-0.2.8     	v0.2.8     
solr-operator     	tjp      	1       	2021-03-24 11:04:56.128582 -0600 MDT	deployed	solr-operator-0.2.8     	v0.2.8     
zookeeper-operator	tjp      	1       	2021-03-24 10:54:55.174334 -0600 MDT	deployed	zookeeper-operator-0.2.9	0.2.9      
```
If that's done as part of another PR or only during the release, please disregard ^^

HoustonPutman:
> Should this PR change the chart version in helm/solr-operator/Chart.yaml to 0.3.0?

Yes. We should probably treat the version as the same as Solr, where what's in the branch represents the next release.

Will put a different PR for that.

thelabdude (approved):
This looks great! Very thorough. I was able to upgrade from 0.2.8 -> a dev build from this PR on GKE.

PR #242, opened 2021-03-25
Add better versioning control in project.

HoustonPutman:
Part of #243

Will allow us to have one location of the "current" version of a branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,17 @@ CRD_OPTIONS ?= "crd"
 
 # Image URL to use all building/pushing image targets
 NAME ?= solr-operator
-NAMESPACE ?= apache/
-IMG = $(NAMESPACE)$(NAME)
-VERSION ?= $(or $(shell git describe --tags HEAD),"latest")
+REPOSITORY ?= $(or $(NAMESPACE:%/=%), "apache")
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -20,9 +28,13 @@ endif
 
 all: generate
 
+.PHONY: version
```
madrob, line 31: isn't tag also PHONY?

version/version.go
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package version
+
+var (
+	// Version information for the Solr Operator
+	Version       string = "v0.3.0"
+	VersionSuffix string = "prerelease"
+	BuildTime     string
```
madrob, line 24: We don't want this because it means that builds are not reproducible. There may be other reasons that the build is not reproducible, but this we shouldn't add to it if we can help it.

HoustonPutman: Maybe we bake this in during a release, therefore it would be reproducible? Rename it "Release Date" or something

HoustonPutman: I'm actually kind of lukewarm on this, the only binary releases we are doing are docker images, which have questionable reproducibility anyways.

PR #247, opened 2021-03-29
[lucene-]solr-operator change in test-upgrade.sh

cpoerschke:
Both https://raw.githubusercontent.com/apache/lucene-solr-operator/v0.2.8/helm/solr-operator/crds/crds.yaml and https://raw.githubusercontent.com/apache/solr-operator/v0.2.8/helm/solr-operator/crds/crds.yaml seem to work in-browser but the latter might be clearer?


HoustonPutman:
Good find. I wrote that before I created https://solr.apache.org/operator/downloads/crds/v0.2.8/all.yaml.

I'll go through and update to this new one.

PR #249, opened 2021-04-02
Create a Release Wizard for the Solr Operator

### LINKED ISSUE #243: Create a ReleaseWizard for the Solr Operator
This will help Solr PMC members complete a release without any inherent knowledge of the process.

Will be built off the Lucene/Solr release wizard.

Related PRs will be pointed here, all steps that are necessary in creating the release wizard.

HoustonPutman:
Resolves #243 

This is an all-encompassing release wizard that does not require previous knowledge from the release manager.

Uses the scripts introduced in #248


HoustonPutman:
Added you as a reviewer @janhoy , but I know this is massive and don't expect you to go through all of it.

As you can tell, this was copied from the Solr release wizard, then completely reworked for the Solr Operator. (It still uses a lot of the same workflow, but almost all of the logic has changed, unsurprisingly)

Just added you on in case something glaring sticks out at you. I have tested a lot of this, but need to do a more thorough run through.

janhoy (commented):
I have not gone through every file here, but I'm very interested in knowing what pain points you went through.

Now that the wizard exists in four forks (lucene-solr.git, lucene.git, solr.git and solr-operator.git), I'd be interested in refactoring  so that the core logic can be shared, and perhaps the tool can be shared with other Apache projects as well.

My plan was to
* Pull out project specific python functions into separate file, e.g. `wizard-include/solroperator.py`
* Make the main `releaseWizard.py` look for an include those include files
* Perhaps move some project specific variables etc into `releaseWizard.yaml`
* Start release-wizard as a separate apache project, or a sub project of comDev
* Publish "core" releaseWizard as a "pip-like" python module somewhere?
* Make the wizard core module a dependency of all lucene/solr wizards, i.e. in requirements.txt

After your recent experience, how likely would you say it is to succeed with something like this? I.e. did you have to change many "core" parts, or mostly project-specific functions?

HoustonPutman:
There were surprisingly few pain points overall.

I do think it would be great to make a common library/program. I think for the Operator's purposes, there could be settings fields in the yaml, that would allow for the python program to be completely generic. I don't even think I had to make changes beyond changing settings in the python code, and the branch versioning stuff (which is kind of hacky now, because `main` is the "stable" branch, there is no `v0.*` branch). There is a large gap between Lucene/Solr and the Solr Operator (go vs java, ant vs Makefile), so I think this would certainly be able to work for a lot of other Apache projects, requiring very little changes.

But I think having a library & basic implementation would be perfect. That would allow for more complex project to add integrations with things like gradle/maven/etc, but also allow easier projects to onboard without having to write python code. The only custom code the Solr Operator would need vs Lucene/Solr is branch naming and versions starting with `v`. But maybe we just have variables in the YAML that regex, or fill in branch name templates. Lots of options that don't require code changes.

Sorry if that's too stream-of-consciousness, but overall I think this would be great to put into a separate project so that other projects can incorporate it easier.

janhoy:
Some comments
* You lack a `requirements.txt` and instructions on how to install dpes
* Generated a HTML version of the guide, and a few hickups
  * "Generated by releaseWizard.py vv0.3.0" (double v due to 'v' in  version)
  * "Select JIRA issues to be included" should be "Select issues to be included" - also check sub bullets
  * Step 3.3 "Install dependencies" should be in "prerequisites" section, and not "Create branch" section? But I understand that you need to have a checkout first??
  * I needed to type my password during install of dependencies, the wizard should alert you about this
  * In releaseWizard.yaml line 693, jinja2 moustache are not expanded

That's how far I got this time :)

HoustonPutman:
Thanks for finding those!

For step 3.3, I agree that it's not a great place to have to download dependencies, but the dependencies might (and eventually will) be different depending on the branch you are on. If someone is making last minute fixes to the branch, it is definitely possible to bump up a version of kustomize, controller-gen or kubebuilder. Therefore we definitely want that to be in the loop of "Pull the branch and make sure that it is good". 

The rest are addressed.

I have also added 1 feature to the releaseWizard.py so far, which is pretty small, let a command have it's own ENV set, independent of the other commands it is with. This is necessary for the `In releaseWizard.yaml line 693, jinja2 moustache are not expanded` issue you pointed out. Pretty small change overall though.

sepulworld:
if releaseWizard is being copied between Apache projects, seems like it is a good candidate to become a Github Action library. https://docs.github.com/en/actions/creating-actions/about-actions

hack/release/wizard/releaseWizard.yaml
```diff
@@ -690,45 +697,31 @@ groups:
...
-      dist_path: '{{ [dist_file_path, dist_folder] | path_join }}'
+      dist_path: '{{ [dist_file_path, build_rc.rc_folder] | path_join }}'
       docker_image: 'apache/solr-operator:{{ release_version }}-rc{{ rc_number }}'
     commands: !Commands
       root_folder: '{{ git_checkout_folder }}'
       commands_text: Here we'll smoke test the release by 'downloading' the artifacts, running the tests, validating GPG signatures etc.
       commands:
       - !Command
-        cmd: ./hack/release/smoke_test/smoke_test.sh -v "{{ release_version }}" -s "{{ build_rc.git_rev | default("<git_rev>", True) | truncate(7) }}" -i "{{ docker_image }}" -l "{{ dist_path }}"
+        cmd: ./hack/release/smoke_test/smoke_test.sh -v "{{ release_version }}" -s "{{ build_rc.git_sha }}" -i "{{ docker_image }}" -l "{{ dist_path }}"
```
janhoy, line 724: The | `default()` clause is there for the export-to-HTML to work, i.e. it will output `<git_rev>` if git_sha is not yet recorded, thus you can generate a generic HTML checklist not tied to a specific release.

janhoy:
> For step 3.3, I agree that it's not a great place ... Therefore we definitely want that to be in the loop of "Pull the branch and make sure that it is good".

Agree, it has to be there, and as long as it is called out, that's fine

> I have also added 1 feature to the releaseWizard.py so far, which is pretty small, let a command have it's own ENV set, independent of the other commands it is with. This is necessary for the `In releaseWizard.yaml line 693, jinja2 moustache are not expanded` issue you pointed out. Pretty small change overall though.

I can see the use of custom ENV per command, yep.

hack/release/wizard/releaseWizard.yaml
HoustonPutman: ahhhh ok, I was very curious about that part. Makes sense. I will use the `default()` for all the stored variables I use throughout.

HoustonPutman:
> if releaseWizard is being copied between Apache projects, seems like it is a good candidate to become a Github Action library. https://docs.github.com/en/actions/creating-actions/about-actions

@sepulworld, unfortunately this is a very manual process, so it'll need to stay a script that you run locally. It takes like a week at least, to go through the whole thing 😬

anshumg (commented):
This is awesome. I didn't dive really deep and look into all of it, but partially reviewed it and this looks great. Thanks @HoustonPutman !

PR #253, opened 2021-04-08
Allow users to supply ACL credentials for provided ZK ensembles to configure Solr to set ACLs on znodes

### LINKED ISSUE #252: Provided Zookeeper config should also allow setting SOLR_ZK_CREDS_AND_ACLS for Solr
Currently, only the `zookeeperRef.connectionInfo` struct allows setting up ZK ACLs via the `SOLR_ZK_CREDS_AND_ACLS` env var. I think users may want Solr to use ZK ACLs for the `provided` ZK as well.

A quick work-around would be to add the env vars for the main container, but those don't get passed along to the `setup-zk` initContainer (which bootstraps security.json). This results in security.json not having ACLs set.

However, even if we fix this issue in the operator code, the Zookeeper operator configures ZK to not enforce ACLs, see:
https://github.com/pravega/zookeeper-operator/issues/316. Thus, until that issue is fixed, if you need ZK ACLs, then you're encouraged to not use the Zookeeper operator and instead deploy ZK using Helm or some other means.

thelabdude:
This fixes #252 .

In addition to unit tests, I added the following config to my SolrCloud CRD YAML to enable basic auth and configure my provided ZK ensemble with ACLs:
```
  solrSecurity:
    authenticationType: Basic
   ...
  zookeeperRef:
    provided:
      acl:
        secret: dev-zk-admin-acl
        passwordKey: password
        usernameKey: username
      readOnlyAcl:
        secret: dev-zk-readonly-acl
        passwordKey: password
        usernameKey: username
```
Verified the `/security.json` was bootstrapped with the correct ACLs.


controllers/util/prometheus_exporter_util.go
```diff
@@ -103,7 +103,7 @@ func GenerateSolrPrometheusExporterDeployment(solrPrometheusExporter *solr.SolrP
 		exporterArgs = append(exporterArgs, "-z", solrConnectionInfo.CloudZkConnnectionInfo.ZkConnectionString())
 
 		// Add ACL information, if given, through Env Vars
-		if hasACLs, aclEnvs := AddACLsToEnv(solrConnectionInfo.CloudZkConnnectionInfo); hasACLs {
+		if hasACLs, aclEnvs := AddACLsToEnv(solrConnectionInfo.CloudZkConnnectionInfo.AllACL, solrConnectionInfo.CloudZkConnnectionInfo.ReadOnlyACL); hasACLs {
```
HoustonPutman, line 106: Do we want to get this out of the SolrCloud object if they do a SolrCloudReference by name in the **same** namespace? Not sure we do, and could be a fix for the future if so.

api/v1beta1/solrcloud_types.go
```diff
@@ -556,6 +556,17 @@ func (ref *ZookeeperRef) withDefaults() (changed bool) {
...
 
+func (ref *ZookeeperRef) GetACLs() (allACL *ZookeeperACL, readOnlyACL *ZookeeperACL) {
+	if ref.ConnectionInfo != nil {
+		allACL = ref.ConnectionInfo.AllACL
+		readOnlyACL = ref.ConnectionInfo.ReadOnlyACL
+	} else if ref.ProvidedZookeeper != nil {
+		allACL = ref.ProvidedZookeeper.AllACL
+		readOnlyACL = ref.ProvidedZookeeper.ReadOnlyACL
+	}
+	return
```
HoustonPutman, line 567: very cool, did not know about this syntax

controllers/util/zk_util.go
```diff
@@ -85,6 +85,10 @@ func GenerateZookeeperCluster(solrCloud *solr.SolrCloud, zkSpec *solr.ZookeeperS
 		zkCluster.Spec.Pod.NodeSelector = zkSpec.ZookeeperPod.NodeSelector
 	}
 
+	if zkSpec.ZookeeperPod.Env != nil {
+		zkCluster.Spec.Pod.Env = zkSpec.ZookeeperPod.Env
```
HoustonPutman, line 89: Think you need to add this field to `CopyZookeeperClusterFields`

HoustonPutman (changes requested):
Awesome addition and especially with tests AND docs 😉 

Just a few minor comments

controllers/util/prometheus_exporter_util.go
HoustonPutman: This is definitely a change to make separately. Since it has very little to do with the scope of this PR.

thelabdude: Just so I'm clear ... no change needed here for this PR, correct?

HoustonPutman: Upon further reflection, yes

controllers/util/zk_util.go
thelabdude: Good catch! Thanks ... always seem to forget about these. I added a basic test for this but only looking at the env ... maybe I should bite the bullet and test all of those fields?

HoustonPutman: Nah, there's already another issue that relates to that. We can tackle it separately and handle all objects at the same time. (They all have the same issue)

HoustonPutman (approved):
I think this is good to go! Great work as always

PR #260, opened 2021-04-19
Various changes needed after testing of the v0.3.0 RC1

### LINKED ISSUE #257: create an ARM64 build of Solr-operator
The container on docker hub today is only available as an amd64 build, can we make this multi-arch and support arm64?

HoustonPutman:
- Add multi-architecture docker images in the release wizard (Resolves #257)
- The `mktemp` command is now used in a way that works with BSD and GNU
- Add licenses to all files in the repo/release
- Fix setup for systems that are not setup to run go code natively
- Make smoke test ports different than the default Solr ports (So it won't fail if a random Solr process is running)
- Fix flakey tests
- Fixed a bug where the wrong error ref was being returned from `verifyTLSSecretConfig` causing improper error handling when reconciling TLS config


thelabdude (approved):
Looks great! ship it ;-)

PR #276, opened 2021-06-14
Introduce a Helm chart for Solr Cloud

### LINKED ISSUE #112: Add a Helm Chart for deploying Solr Clouds
We currently have a Helm Chart to deploy the Solr Operator, but users still have to manually create yaml files for their clouds/collections/etc.

It would be great to have helm charts that allow for users to easily create new SolrClouds (with metrics or collections included possibly). That way creating a solr cloud is two calls maximum (one to install the operator via helm, one to install the cloud via helm).

HoustonPutman:
Resolves #112

This is a fully-featured Solr helm chart that merely templates the available fields in the SolrCloud CRD.

The SolrPrometheusExporter is not included currently, but that is certainly a possibility in the future.

I've tested locally with the local_tutorial, which now uses the helm chart, but I'm sure there are areas where the chart has some bugs.

I'd love some help testing additional configurations, but I do eventually plan on merging this on the week of July 5.


mogul:
I'm very happy to see this! I'd done [something similar on my own](https://github.com/GSA/datagov-brokerpak/tree/main/terraform/provision/solr-crd) to work around the fact that the Terraform k8s provider doesn't (yet) support CRDs; by using a Helm chart I was still able to [manage SolrCloud deployments from Terraform](https://github.com/GSA/datagov-brokerpak/blob/ce85552a1a5f1b4447a0c29ff16380863f024c79/terraform/provision/main.tf#L39-L40).

PR #279, opened 2021-07-06
Ability to provide a cron-schedule for automatic restarts

### LINKED ISSUE #281: Enable scheduled restarts of Solr Clouds
For various reasons, users might want to restart their Solr Clouds on some sort of schedule. This could be for stability, Client Auth cert renewal, etc. It is possible for users to enable this themselves, but it's also something that can be easily supported by the operator without having to write risky code client-side.

The operator should be able to support this by merely setting a time annotation on the Solr pod template, and scheduling a reconcile for when the next scheduled restart should take place.

Using a Kubernetes `CronJob` would likely be overkill, and dealing with images and permissions to modify the statefulSet or SolrCloud would be more painful than including it as an option in the SolrCloud spec.

HoustonPutman:
Resolves #281

### DIFF
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -475,6 +475,10 @@ type SolrUpdateStrategy struct {
 	// Options for Solr Operator Managed rolling updates.
 	// +optional
 	ManagedUpdateOptions ManagedUpdateOptions `json:"managed,omitempty"`
+
+	// Perform a scheduled restart on the given schedule, in CRON format.
+	// +optional
+	RestartSchedule string `json:"restartSchedule,omitempty"`
 }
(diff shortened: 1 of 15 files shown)


thelabdude:
haven't reviewed yet, just wondering if a Go duration string (https://golang.org/pkg/time/#ParseDuration) would work better than a CRON string? Similar to the pod reaper https://github.com/ptagr/pod-reaper ...

HoustonPutman:
Conveniently [this CRON api supports](https://pkg.go.dev/github.com/robfig/cron/v3?utm_source=godoc#hdr-Predefined_schedules) `@every <duration>`!

thelabdude (approved):
Nice unit tests!

PR #292, opened 2021-07-20
Add config options to enable TLS from cert files mounted to a dir on each pod by a cert-manager extension

### LINKED ISSUE #291: Config options to enable TLS from files in a mounted directory instead of from a secret
One emerging pattern in the K8s ecosystem I've noticed is to have a cert-manager extension that issues a unique TLS certificate for each pod. Typically how this works is the cert-manager extension watches for specific annotations on the pod spec (in the STS) and then handles the certificate request per pod; once the cert is issued, the extension mounts the cert files to a directory on the pod. The files could come from a unique secret per pod or via a CSI driver that loads the files to a specified volume mount.

The current TLS solution in the Solr operator looks for a shared secret containing the keystore and supporting files; all pods share the same cert. With the cert-manager extension approach described above, the secret containing the TLS files for each pod is unique. Hence, the current TLS options in the operator won't work. The Solr operator needs to know this automated TLS mounting process is happening so it can configure Solr to use the provided keystore / truststore, as well as enable TLS in various places, such as on the probes as well as setting the `urlScheme` in ZK. The operator also needs to know to send requests to Solr using `https`. 

So I'm thinking of adding a `mountedTLSDir` option to the existing `SolrTLSOptions` struct to indicate to the Solr operator it needs to enable TLS for Solr using the files in the specified directory and not look for a shared secret containing the TLS files.

When `mountedTLSDir` is used, the cert issuing process is opaque to the Solr operator, it just cares about the keystore, truststore, and store password files being in the PKCS12 format and being mounted on each Solr pod. The `mountedTLSDir` will be struct with defaults for the expected file names, e.g. `keystore.p12`, `truststore.p12`, and `keystore-password`. However, we need to be flexible about the names of the various files in the mount dir (although I think we can require the keystore/truststore to be in PKCS12 format)

thelabdude:
Fixes #291 

The tricky part of this PR is the keystore password being in a file, but Solr (and the exporter) have no facility for loading the password from a file (Solr expects the `SOLR_SSL_KEY_STORE_PASSWORD` ENV VAR via the `EnvSSLCredentialProvider`) and the prometheus exporter expects these to come through the `javax.net.ssl.keyStorePassword` and `javax.net.ssl.trustStorePassword` Java system props ...

So for the main Solr pod, I'm leveraging the initdb support in the Docker Solr framework to run a script in `/docker-entrypoint-initdb.d` to export the `SOLR_SSL_KEY_STORE_PASSWORD` and `SOLR_SSL_TRUST_STORE_PASSWORD` env vars by cat'ing out the password file(s) in the mounted dir. That's a bit onerous but the real rub is with the exporter, where it doesn't even work with the  `SOLR_SSL_KEY_STORE_PASSWORD` and `SOLR_SSL_TRUST_STORE_PASSWORD` env vars and expects the Java system properties. So had to resort to wrapping the entrypoint script to add these properties from the password file at runtime.

Added basic unit tests to verify the config and reconcile around these new config options


controllers/solrprometheusexporter_controller.go
```diff
@@ -216,6 +216,11 @@ func (r *SolrPrometheusExporterReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 				}
 			}
 		}
+	} else if prometheusExporter.Spec.SolrReference.SolrTLS != nil && prometheusExporter.Spec.SolrReference.SolrTLS.MountedTLSDir != nil {
+		tlsClientOptions = &util.TLSClientOptions{}
+		tlsClientOptions.TLSOptions = prometheusExporter.Spec.SolrReference.SolrTLS
+		// TODO: How to get this from the SolrCloud reference?
```
thelabdude, line 222: Would appreciate some advice on how best to handle this? Maybe just add the same:
```
	// +optional
	BusyBoxImage *ContainerImage `json:"busyBoxImage,omitempty"`
```
For the Prometheus exporter CRD?

HoustonPutman: Yeah, no other option 😢 

Maybe in the description of the field we can say it is only used when mountedTLS is enabled?

thelabdude: SGTM ;-) Will update shortly

controllers/util/solr_util.go
```diff
@@ -101,7 +101,7 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 	defaultMode := int32(420)
 
 	probeScheme := corev1.URISchemeHTTP
-	if solrCloud.Spec.SolrTLS != nil {
+	if solrCloud.UrlScheme() == "https" {
```
HoustonPutman, line 104: this is going to conflict with the Ingress TLS Termination PR. When it does, just change this to `SolrCloud.UrlScheme(false)` (getting url scheme internal in the cluster)

controllers/util/solr_util.go
```diff
@@ -1390,12 +1434,13 @@ func generateSecurityJson(solrCloud *solr.SolrCloud) map[string][]byte {
           %s,
           { "name": "k8s-status", "role":"k8s", "collection": null, "path":"/admin/collections" },
           { "name": "k8s-metrics", "role":"k8s", "collection": null, "path":"/admin/metrics" },
+          { "name": "k8s-zk", "role":"k8s", "collection": null, "path":"/admin/zookeeper/status" },
```
HoustonPutman, line 1437: You mind if we fix these in a separate small PR?

api/v1beta1/solrcloud_types.go
```diff
@@ -1106,6 +1141,47 @@ type SolrTLSOptions struct {
 	// Opt-in flag to restart Solr pods after TLS secret updates, such as if the cert is renewed; default is false.
 	// +optional
 	RestartOnTLSSecretUpdate bool `json:"restartOnTLSSecretUpdate,omitempty"`
+
+	// Used to specify a path where the keystore, truststore, and password files are mounted by an external agent or CSI driver.
+	// This option is typically used with a `restartSchedule` to restart Solr pods before the mounted TLS cert expires.
```
HoustonPutman, line 1146:
```suggestion
	// This option is typically used with `spec.updateStrategy.restartSchedule` to restart Solr pods before the mounted TLS cert expires.
```

HoustonPutman (commented):
Small comments, looks good so far though!

Awaiting some wonderful additions to the TLS tests 😅

controllers/util/solr_util.go
thelabdude: this is an oops actually, sorry 'bout that (some cruft from an earlier approach), will revert back to what was there

controllers/util/solr_util.go
thelabdude: definitely, thanks for catching ... just a quick hack I made for testing on 8.9 but should not be in this PR

HoustonPutman (approved):

HoustonPutman (approved):
Good to go!

PR #293, opened 2021-07-20
Add option for Ingress TLS termination for SolrCloud.

### LINKED ISSUE #268: Allow specifying the name of a TLS secret for the Ingress to support TLS termination at the Ingress
Currently, the operator doesn't support TLS termination at the Ingress. Users can configure TLS all the way through from Ingress to Solr pods and between them. This would be an easy enhancement, but requires adding a field to the `solrAddressability.external` struct to hold the TLS secret name so that we can configure the Ingess with:

```
spec:
  tls:
    - hosts:
      - foo.bar.com
      # This assumes tls-secret exists and the SSL
      # certificate contains a CN for foo.bar.com
      secretName: tls-secret
```
see: https://kubernetes.github.io/ingress-nginx/examples/tls-termination/

In the meantime, users will need to create their own Ingress instead of using the one created by the operator.

HoustonPutman:
Resolves #268 

This gives the user to enable TLS Termination in the Ingress for all external Solr traffic.
Users then have the ability to limit what endpoints are available outside the cluster through `hideNodes` and `hideCommon`.

This cannot be used with `useExternalAddress`, as Solr cannot be listening on HTTP and make internal communication through HTTPS.


thelabdude (approved):
This looks great. Docs are clear and concise.

HoustonPutman:
So it looks like Nginx does support tls-passthrough with different TLS options. So while the current PR does work, we will probably want to change a few things to support the ingress TLS termination with Solr TLS enabled.

Issue #289, opened 2021-07-16
Boostrapped security.json config doesn't work with the Prometheus exporter in Solr 8.9.0 due to new endpoint


thelabdude:
Solr 8.9.0 has a new metrics endpoint being hit: `/admin/zookeeper/status`. The boostrapped `security.json` (created by the Solr operator when basic auth is enabled) doesn't wire-up access for this new endpoint for the `k8s` role.

```
ERROR - 2021-07-16 22:57:47.229; org.apache.solr.prometheus.collector.SchedulerMetricsCollector; Error occurred during metrics collection => org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at https://dev-solrcloud-0.dev-solrcloud-headless.dev:8983/solr: Forbidden

org.apache.solr.client.solrj.impl.HttpSolrClient$RemoteSolrException: Error from server at https://dev-solrcloud-0.dev-solrcloud-headless.dev:8983/solr: Forbidden

request: https://dev-solrcloud-0.dev-solrcloud-headless.dev:8983/solr/admin/zookeeper/status?wt=json&version=2.2
```

Work-around is to add the following mapping to your security.json file:
```
      {
        "name": "k8s-zk",
        "role": "k8s",
        "collection": null,
        "path": "/admin/zookeeper/status"
      },
```

thelabdude:
Will fix this when I fix #274

PR #299, opened 2021-07-26
Fix initial security.json rbap rules

### LINKED ISSUE #274: Bootstrapped security.json incorrectly assigns the "all" permission to the "users" role due to ordering of rules
The `security.json` that the Solr operator creates during security boostrapping when using:
```
spec:
  ...
  solrSecurity:
    authenticationType: Basic
```
Grants the `all` permission to the `users` role.

I misinterpreted this permission to mean all endpoints unless specifically overridden by another rule, which is how it works but seems to be order dependent. If this "all" permission is listed last, then we get the behavior I intended. But if it's listed before the more restrictive rules, then it takes precedence and allows broader access than I intended for the default authz rules. Upon further reflection, I think the operator should not assign the `users` role to the `all` permission at all, so the change looks like:
```
diff --git a/controllers/util/solr_util.go b/controllers/util/solr_util.go
index c8f3b05..fbc9d0e 100644
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -1387,7 +1387,7 @@ func generateSecurityJson(solrCloud *solr.SolrCloud) map[string][]byte {
           { "name": "k8s-status", "role":"k8s", "collection": null, "path":"/admin/collections" },
           { "name": "k8s-metrics", "role":"k8s", "collection": null, "path":"/admin/metrics" },
           { "name": "k8s-ping", "role":"k8s", "collection": "*", "path":"/admin/ping" },
-          { "name": "all", "role":["admin","users"] },
+          { "name": "all", "role":["admin"] },
           { "name": "read", "role":["admin","users"] },
           { "name": "update", "role":["admin"] },
           { "name": "security-read", "role": "admin"},
```

The work-around for users that have already bootstrapped a SolrCloud with basic-auth enabled is to manually update their `security.json` to remove `users` from the `all` role. You can do this using the zkcli.sh script to update the security.json directly:
```
./zkcli.sh -zkhost $ZK_HOST -cmd putfile /security.json $SOME_LOCAL_PATH/security.json 
```

thelabdude:
Fixes #274 and #289 

Very minor change to remove the `users` role from the `all` permission, move the `all` permission to the last index in the json file, and add a new rule for the `/admin/zookeeper/status` path needed for updates to the exporter in 8.9 (see #289 )

Manual integration-style testing required:
Create a SolrCloud running Apache Solr 8.9.0, created a collection, and then tried to index some docs as the `solr` user, which now fails with a 403 ~ Unauthorized. Indexing as the `admin` user works. The `solr` user can log in to the Admin UI but can only query collections. 

Also verified the Prometheus exporter (also running 8.9) can now retrieve metrics when basic auth is enabled.

### DIFF
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -1341,12 +1341,13 @@ func generateSecurityJson(solrCloud *solr.SolrCloud) map[string][]byte {
           %s,
           { "name": "k8s-status", "role":"k8s", "collection": null, "path":"/admin/collections" },
           { "name": "k8s-metrics", "role":"k8s", "collection": null, "path":"/admin/metrics" },
+          { "name": "k8s-zk", "role":"k8s", "collection": null, "path":"/admin/zookeeper/status" },
           { "name": "k8s-ping", "role":"k8s", "collection": "*", "path":"/admin/ping" },
-          { "name": "all", "role":["admin","users"] },
           { "name": "read", "role":["admin","users"] },
           { "name": "update", "role":["admin"] },
           { "name": "security-read", "role": "admin"},
-          { "name": "security-edit", "role": "admin"}
+          { "name": "security-edit", "role": "admin"},
+          { "name": "all", "role":["admin"] }
         ]
       }
     }`, blockUnknown, credentialsJson, username, probeAuthz)
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -1345,8 +1345,8 @@ func generateSecurityJson(solrCloud *solr.SolrCloud) map[string][]byte {
           { "name": "k8s-ping", "role":"k8s", "collection": "*", "path":"/admin/ping" },
           { "name": "read", "role":["admin","users"] },
           { "name": "update", "role":["admin"] },
-          { "name": "security-read", "role": "admin"},
-          { "name": "security-edit", "role": "admin"},
+          { "name": "security-read", "role": ["admin"] },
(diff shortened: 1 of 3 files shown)


HoustonPutman:
+1 to the change, but we need to update the documentation to use the new rules and also explain that the default `solr` user will not be able to add/update/delete docs.

HoustonPutman:
Also an addition in the changelog of the Solr Operator helm chart would be nice! You can follow the templates of the other entries. Possible change _kind_ values are found [here](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations), `fixed` or `security` probably work best. Also if you think it's a serious security concern being addressed, you could always change `artifacthub.io/containsSecurityUpdates` to true.

thelabdude:
> Also an addition in the changelog of the Solr Operator helm chart would be nice! You can follow the templates of the other entries. Possible change _kind_ values are found [here](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations), `fixed` or `security` probably work best. Also if you think it's a serious security concern being addressed, you could always change `artifacthub.io/containsSecurityUpdates` to true.

I don't think it's a serious security issue (willing to be convinced though) ... mainly the `solr` user ended up getting more access than I intended but users still have to give out the `solr` user's credentials, which presumes they've reviewed that account's access privileges before handing it out willy-nilly ;-)

HoustonPutman (approved):
Looks good!

PR #309, opened 2021-08-17
Refactor TLS setup code into a separate file and consolidate logic between the StatefulSet and exporter Deployment

### LINKED ISSUE #308: Refactor TLS code
The code that enables TLS for Solr and Prometheus exporter pods has evolved over the past several months, and still has some additional changes we want to add (e.g. https://github.com/apache/solr-operator/issues/300). Let's clean-up the TLS code before adding more features as it's become a bit hard to follow the logic and is infused across too many places. This will also help reduce the size and complexity of the `GenerateStatefulSet` method in `solr_util.go`.

thelabdude:
Fixes #308 

Consolidated most of the TLS related utility code into a new file in the `util` package: `solr_tls_util.go`

All the TLS related config (env vars, annotations, initContainers, volumes/mounts) get appended to the config for a `StatefulSet` or `Deployment` in one step at the end vs. being infused at various places in `GenerateStatefulSet` and `GenerateSolrPrometheusExporterDeployment`. I like this approach because most of the TLS config is additive. There are a few places where we have to be careful about overriding previous settings (such as the `entrypoint` for the exporter when using the `mountedServerTLSDir` but that's only in a few small places).

There's also a lot of common logic for configuring the StatefulSet (for SolrClouds) and Deployment (for exporter), so I refactored the reusable parts into their own method.

### DIFF
--- a/controllers/controller_utils_test.go
+++ b/controllers/controller_utils_test.go
@@ -190,14 +190,6 @@ func verifyUserSuppliedTLSConfig(t *testing.T, tls *solr.SolrTLSOptions, expecte
 	assert.Equal(t, expectedKeystorePasswordSecretKey, tls.KeyStorePasswordSecret.Key)
 	assert.Equal(t, expectedTlsSecretName, tls.PKCS12Secret.Name)
 	assert.Equal(t, "keystore.p12", tls.PKCS12Secret.Key)
-
-	// is there a separate truststore?
-	expectedTrustStorePath := ""
-	if tls.TrustStoreSecret != nil {
-		expectedTrustStorePath = util.DefaultTrustStorePath + "/" + tls.TrustStoreSecret.Key
-	}
-
-	expectTLSEnvVars(t, util.TLSEnvVars(tls, needsPkcs12InitContainer), expectedKeystorePasswordSecretName, expectedKeystorePasswordSecretKey, needsPkcs12InitContainer, expectedTrustStorePath)
 }
 
 func createTLSOptions(tlsSecretName string, keystorePassKey string, restartOnTLSSecretUpdate bool) *solr.SolrTLSOptions {
@@ -376,7 +368,7 @@ func expectMountedTLSDirEnvVars(t *testing.T, envVars []corev1.EnvVar) {
 	envVars = filterVarsByName(envVars, func(n string) bool {
 		return strings.HasPrefix(n, "SOLR_SSL_")
 	})
-	assert.True(t, len(envVars) == 7)
+	assert.Equal(t, 7, len(envVars), "expected SOLR_SSL related env vars not found")
 
 	expectedKeystorePath := "/mounted-tls-dir/keystore.p12"
 	expectedTruststorePath := "/mounted-tls-dir/truststore.p12"
--- a/controllers/solrcloud_controller.go
+++ b/controllers/solrcloud_controller.go
@@ -215,7 +215,7 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			// if there's a user-provided config, it must have one of the expected keys
 			if !hasLogXml && !hasSolrXml {
 				// TODO: Create event for the CRD.
-				return requeueOrNot, fmt.Errorf("User provided ConfigMap %s must have one of 'solr.xml' and/or 'log4j2.xml'",
+				return requeueOrNot, fmt.Errorf("user provided ConfigMap %s must have one of 'solr.xml' and/or 'log4j2.xml'",
 					providedConfigMapName)
 			}
 
@@ -223,7 +223,7 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 				// make sure the user-provided solr.xml is valid
 				if !strings.Contains(solrXml, "${hostPort:") {
 					return requeueOrNot,
-						fmt.Errorf("Custom solr.xml in ConfigMap %s must contain a placeholder for the 'hostPort' variable, such as <int name=\"hostPort\">${hostPort:80}</int>",
+						fmt.Errorf("custom solr.xml in ConfigMap %s must contain a placeholder for the 'hostPort' variable, such as <int name=\"hostPort\">${hostPort:80}</int>",
 							providedConfigMapName)
 				}
 				// stored in the pod spec annotations on the statefulset so that we get a restart when solr.xml changes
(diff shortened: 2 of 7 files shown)


controllers/controller_utils_test.go
```diff
@@ -190,14 +190,6 @@ func verifyUserSuppliedTLSConfig(t *testing.T, tls *solr.SolrTLSOptions, expecte
...
 	assert.Equal(t, expectedTlsSecretName, tls.PKCS12Secret.Name)
 	assert.Equal(t, "keystore.p12", tls.PKCS12Secret.Key)
-
-	// is there a separate truststore?
-	expectedTrustStorePath := ""
-	if tls.TrustStoreSecret != nil {
-		expectedTrustStorePath = util.DefaultTrustStorePath + "/" + tls.TrustStoreSecret.Key
-	}
-
-	expectTLSEnvVars(t, util.TLSEnvVars(tls, needsPkcs12InitContainer), expectedKeystorePasswordSecretName, expectedKeystorePasswordSecretKey, needsPkcs12InitContainer, expectedTrustStorePath)
```
thelabdude, line 200: We don't need to call `expectTLSEnvVars` in this part of the test code as it already gets called after reconciliation, see `expectTLSConfigOnPodTemplate` and `expectMountedTLSDirConfigOnPodTemplate`. Removing this code allows us to hide the `TLSEnvVars` in the util package, it doesn't need to be exposed here.

controllers/solrcloud_controller.go
```diff
@@ -215,15 +215,15 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			// if there's a user-provided config, it must have one of the expected keys
 			if !hasLogXml && !hasSolrXml {
 				// TODO: Create event for the CRD.
-				return requeueOrNot, fmt.Errorf("User provided ConfigMap %s must have one of 'solr.xml' and/or 'log4j2.xml'",
+				return requeueOrNot, fmt.Errorf("user provided ConfigMap %s must have one of 'solr.xml' and/or 'log4j2.xml'",
```
thelabdude, line 218: Just cleaning up some IDE nits here ... doesn't need to be in this PR but shouldn't hurt either ;-)

controllers/solrcloud_controller_tls_test.go
```diff
@@ -126,7 +126,6 @@ func TestMountedTLSDir(t *testing.T) {
 	mountedDir := &solr.MountedTLSDirectory{}
 	mountedDir.Path = "/mounted-tls-dir"
 	instance.Spec.SolrTLS = &solr.SolrTLSOptions{MountedServerTLSDir: mountedDir, CheckPeerName: true, ClientAuth: "Need", VerifyClientHostname: true}
-	expectMountedTLSDirEnvVars(t, util.TLSEnvVars(instance.Spec.SolrTLS, false))
```
thelabdude, line 129: the TLS env vars will get checked in the `expectMountedTLSDirConfigOnPodTemplate` after reconciliation, so no need to do here ... same comment for change below this too ...

controllers/util/solr_util.go
```diff
@@ -601,6 +552,11 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 		}
 	}
 
+	// Enrich the StatefulSet config to enable TLS on Solr pods if needed
+	if tls != nil {
+		tls.enableTLSOnSolrCloudStatefulSet(stateful)
```
thelabdude, line 557: This is the crux of the design I took for this refactoring task. Basically, take a configured StatefulSet and then enrich it with all the TLS things in one go vs. spread throughout the method as it was before.

controllers/solrcloud_controller.go
```diff
@@ -364,51 +364,57 @@ func (r *SolrCloudReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-			if instance.Spec.SolrTLS.RestartOnTLSSecretUpdate {
-				if tlsCertBytes, ok := foundTLSSecret.Data[util.TLSCertKey]; ok {
-					tlsCertMd5 = fmt.Sprintf("%x", md5.Sum(tlsCertBytes))
-				} else {
-					return requeueOrNot, fmt.Errorf("%s key not found in TLS secret %s, cannot watch for updates to"+
-						" the cert without this data but 'solrTLS.restartOnTLSSecretUpdate' is enabled!",
-						util.TLSCertKey, foundTLSSecret.Name)
+	var tls *util.TLSConfig
+	if !blockReconciliationOfStatefulSet && instance.Spec.SolrTLS != nil {
+		tls = &util.TLSConfig{}
```
thelabdude, line 370: The `TLSConfig` struct allows us to hold the `TLSOptions` that from from the user config as well as additional config info determined during reconciliation, such as the MD5 hash of the cert. Not married to the name of this struct ... could be `TLSOptionsAndReconciledVars`

controllers/solrprometheusexporter_controller.go
```diff
@@ -197,28 +197,30 @@ func (r *SolrPrometheusExporterReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 					prometheusExporter.Spec.SolrReference.SolrTLS.KeyStorePasswordSecret.Key, keyStorePasswordSecret.Name)
 			}
 
-			tlsClientOptions = &util.TLSClientOptions{}
```
thelabdude, line 200: The `TLSClientOptions` name was misleading once I repurposed this struct to be used by the exporter and the StatefulSet code

controllers/util/prometheus_exporter_util.go
```diff
@@ -46,17 +45,9 @@ type SolrConnectionInfo struct {
 	StandaloneAddress      string
 }
 
-// Used internally to capture config needed to provided Solr client apps like the exporter
-// with config needed to call TLS enabled Solr pods
-type TLSClientOptions struct {
```
thelabdude, line 51: Replaced by `TLSConfig` in `solr_tls_util.go`

controllers/util/solr_util.go
HoustonPutman: Yeah this is great. Hopefully we can abstract out other parts of this awful method and make it nice and clean.

controllers/solrcloud_controller.go
HoustonPutman: `TLSConfig` is fine with me

HoustonPutman (approved):
I really like the new design, this looks good to me!

PR #312, opened 2021-08-25
Configuration options to support an additional client TLS cert in addition to the server certificate

### LINKED ISSUE #300: Config should allow using a separate client cert for mTLS
Currently, the operator TLS support only sets the server auth cert env vars, e.g.
```
SOLR_SSL_KEY_STORE
SOLR_SSL_KEY_STORE_PASSWORD
SOLR_SSL_TRUST_STORE
SOLR_SSL_TRUST_STORE_PASSWORD
```
For mTLS, users may want to configure a separate keystore / truststore for a separate client cert using:
```
SOLR_SSL_CLIENT_KEY_STORE=
SOLR_SSL_CLIENT_KEY_STORE_PASSWORD=
SOLR_SSL_CLIENT_TRUST_STORE=
SOLR_SSL_CLIENT_TRUST_STORE_PASSWORD=
```

thelabdude:
Fixes #300 

Solr supports using a different TLS certificate (keystore & truststore) for making client requests (to other Solr pods) than the server certificate. This PR enhances the TLS configuration options to support this additional client certificate.

I also cleaned up the Prometheus exporter TLS code to allow only using a truststore as the exporter doesn't need a keystore unless mTLS is enabled since the exporter only exposes HTTP endpoints. In other words, the exporter config should support users only providing a truststore and not require a keystore (for flexibility).

I also renamed `mountedServerTLSDir` to `mountedTLSDir` since I introduced a `spec.SolrClientTLS` to the `SolrCloudSpec` struct; this is a new struct added for v0.4.0, so no back-compat concerns with this rename. When I originally created `mountedServerTLSDir`, I was thinking I would have a `mountedClientTLSDir` as well for this feature but I think having an additional top-level `SolrClientTLS` on the `SolrCloudSpec` is better since we don't want to introduce `ClientKeystore` + `ClientKeystorePassword` + `ClientTruststore` ... for the non-mounted dir option. Also, `mountedTLSDir` works better for the Prometheus exporter CRD since it's only a TLS client and doesn't serve HTTPs traffic.

It probably doesn't make sense to review the "diff" for `controllers/util/solr_tls_util.go` and instead just go over the entire file since I had to make so many changes there to accommodate these changes.

Here's an example SolrCloud config to put these changes into context:
```
  solrTLS:
    clientAuth: Need
    checkPeerName: true
    verifyClientHostname: true
    restartOnTLSSecretUpdate: true
    keyStorePasswordSecret:
      name: pkcs12-password-secret
      key: password-key
    pkcs12Secret:
      name: dev-selfsigned-cert-tls
      key: keystore.p12
    trustStoreSecret:
      name: dev-selfsigned-cert-tls
      key: truststore.p12

  solrClientTLS:
    restartOnTLSSecretUpdate: true
    keyStorePasswordSecret:
      name: pkcs12-password-secret
      key: password-key
    pkcs12Secret:
      name: dev-selfsigned-client-cert-tls
      key: keystore.p12
    trustStoreSecret:
      name: dev-selfsigned-client-cert-tls
      key: truststore.p12
```


controllers/controller_utils_test.go
```diff
@@ -184,46 +182,6 @@ func expectDeployment(t *testing.T, g *gomega.GomegaWithT, requests chan reconci
 	return deploy
 }
 
-func verifyUserSuppliedTLSConfig(t *testing.T, tls *solr.SolrTLSOptions, expectedKeystorePasswordSecretName string, expectedKeystorePasswordSecretKey string, expectedTlsSecretName string, needsPkcs12InitContainer bool) {
```
thelabdude, line 187: Moved these TLS specific test helper functions to `solrcloud_controller_tls_test` as they are only used by TLS tests and now seem out of place in this file

controllers/controller_utils_test.go
```diff
@@ -279,170 +237,6 @@ func expectInitContainer(t *testing.T, podTemplate *corev1.PodTemplateSpec, expN
 	return expInitContainer
 }
 
-// Ensures all the TLS env vars, volume mounts and initContainers are setup for the PodTemplateSpec
```
thelabdude, line 282: Moved these TLS specific test helper functions to `solrcloud_controller_tls_test` as they are only used by TLS tests and now seem out of place in this file

controllers/solrcloud_controller.go
```diff
@@ -960,80 +950,166 @@ func (r *SolrCloudReconciler) indexAndWatchForTLSSecret(mgr ctrl.Manager, ctrlBu
...
+	}
+
+	return ctrlBuilder.Watches(
+		&source.Kind{Type: &corev1.Secret{}},
+		r.findSolrCloudByFieldValueFunc(field),
 		builder.WithPredicates(predicate.ResourceVersionChangedPredicate{})), nil
 }
 
+func (r *SolrCloudReconciler) findSolrCloudByFieldValueFunc(field string) *handler.EnqueueRequestsFromMapFunc {
+	return &handler.EnqueueRequestsFromMapFunc{
```
thelabdude, line 978: Refactored into this reusable function as only the field name varies for 3 different indexAndWatch funcs.

controllers/solrcloud_controller.go
```diff
@@ -960,80 +950,160 @@ func (r *SolrCloudReconciler) indexAndWatchForTLSSecret(mgr ctrl.Manager, ctrlBu
...
+		if tls.ClientConfig != nil {
+			if tls.ClientConfig.Options.PKCS12Secret == nil {
+				// cannot mix options with the client cert, if the server cert comes from a secret, so too must the client, not a mountedTLSDir
+				return nil, fmt.Errorf("invalid TLS config, the 'solrClientTLS.pkcs12Secret' option is required when using a secret for server cert")
+			}
+
+			clientOpts := tls.ClientConfig.Options
+			// shouldn't configure a client cert if it's the same as the server cert
+			if clientOpts.PKCS12Secret == tls.ServerConfig.Options.PKCS12Secret {
+				return nil, fmt.Errorf("invalid TLS config, the 'solrClientTLS.pkcs12Secret' option should not be the same as the 'solrTLS.pkcs12Secret'")
```
HoustonPutman, line 1062: Will this actually result in an error when running Solr?

controllers/solrprometheusexporter_controller.go
```diff
@@ -434,53 +458,133 @@ func (r *SolrPrometheusExporterReconciler) buildSecretWatch(secretField string,
...
-	tls.InitContainerImage = prometheusExporter.Spec.BusyBoxImage
-	tls.Options = opts
+// Reconcile the various options for configuring TLS for the exporter
+// The exporter is a client to Solr pods, so can either just have a truststore so it trusts Solr certs
+// Or it can have its own client auth cert when Solr mTLS is required
+func (r *SolrPrometheusExporterReconciler) reconcileTLSConfig(prometheusExporter *solrv1beta1.SolrPrometheusExporter) (*util.TLSCerts, error) {
+	tls := util.TLSCertsForExporter(prometheusExporter)
+	opts := tls.ClientConfig.Options
 
 	if opts.PKCS12Secret != nil {
```
HoustonPutman, line 468: Looks like there's a lot of duplicated code between `SolrPrometheusExport.solrTLS`, `SolrCloud.solrTLS` and `SolrCloud.solrClientTLS`. Could it be one generic method used for all of them, or is that not feasible?

controllers/util/solr_tls_util.go
```diff
@@ -27,171 +27,217 @@ import (
...
-		{
-			Name: "keystore",
+	vols := []corev1.Volume{}
+	mounts := []corev1.VolumeMount{}
+	keystoreSecretName := ""
+
+	opts := tls.Options
+	if opts.PKCS12Secret != nil {
+		keystoreSecretName = opts.PKCS12Secret.Name
+		volName := namePrefix + "keystore"
```
HoustonPutman, line 205: maybe "-keystore"?

controllers/util/solr_tls_util.go
```diff
@@ -27,171 +27,217 @@ import (
...
-		},
+		})
+		mounts = append(mounts, corev1.VolumeMount{Name: volName, ReadOnly: true, MountPath: tls.KeystorePath})
 	}
 
 	// if they're using a different truststore other than the keystore, but don't mount an additional volume
 	// if it's just pointing at the same secret
-	if tls.Options.TrustStoreSecret != nil && tls.Options.TrustStoreSecret.Name != tls.Options.PKCS12Secret.Name {
+	if opts.TrustStoreSecret != nil && opts.TrustStoreSecret.Name != keystoreSecretName {
+		volName := namePrefix + "truststore"
```
HoustonPutman, line 222: maybe "-truststore"?

controllers/util/solr_tls_util.go
```diff
@@ -329,45 +438,34 @@ func (tls *TLSConfig) mountTLSWrapperScriptAndInitContainer(deployment *appsv1.D
...
+	shCmd += " > /docker-entrypoint-initdb.d/export-tls-vars.sh"
 	/*
 	   Init container creates a script like:
 
 	      #!/bin/bash
-	      export SOLR_SSL_KEY_STORE_PASSWORD=`cat $MOUNTED_TLS_DIR/keystore-password`
-	      export SOLR_SSL_TRUST_STORE_PASSWORD=`cat $MOUNTED_TLS_DIR/keystore-password`
+
+	      export SOLR_SSL_KEY_STORE_PASSWORD=`cat $MOUNTED_SERVER_TLS_DIR/keystore-password`
+	      export SOLR_SSL_TRUST_STORE_PASSWORD=`cat $MOUNTED_SERVER_TLS_DIR/keystore-password`
```
HoustonPutman, line 466: truststore-password here and two lines below.

HoustonPutman (commented):
Looks really good! Don't have many comments to make other than small things.

controllers/solrcloud_controller.go
thelabdude: Probably not an issue for Solr but seemed like that would be a mis-configuration / accident by the user since it wouldn't be sensible to use the same cert for the client & server (you don't need a separate client in that case) so I felt raising an error would help draw attention to a mistake.

controllers/solrprometheusexporter_controller.go
thelabdude: yeah, let me take another stab on consolidating the reconcile loop around this, definitely a lot of similarity that can be reused.

controllers/util/solr_tls_util.go
thelabdude: the prefix is either "" or "client-" ... I'll fix it with a helper method on `tls.Options`

HoustonPutman: Oh then this is fine, ignore

controllers/solrprometheusexporter_controller.go
thelabdude: Greatly reduced the amount of code in the TLS reconcile loops for SolrCloud and Exporter in the last commit. Good eye!

PR #318, opened 2021-08-31
Option to watch for updates to the mTLS cert used by the operator to call Solr pods

### LINKED ISSUE #317: Solr operator pod should hot reload the TLS client cert used to send requests to HTTPs enabled Solr pods
If Solr has mTLS configured with ClientAuth `Need` the the Solr operator pod needs a client cert that Solr pods trust. We added support for loading said cert from a secret in v0.3.0, see (#256). The next step is for the Solr operator pod to just hot reload the cert when it gets updated (K8s will handle updating the files on the pod when the secret data updates).

thelabdude:
Fixes #317 

Use `fsnotify` to watch the mTLS `tls.crt` file (mounted from a secret) to hot reload the certificate for the Http client used to make calls to Solr.

For files loaded from a secret, the watched file will be a symlink that gets removed and re-added when the underlying secret changes. Thus, our watcher responds to a REMOVE event on the symlink to reload the certificate. It also supports a WRITE event if the watched file is not a symlink, which is probably not needed right now but may be useful in the future when TLS certs get loaded from a CSI driver or other type mechanism.

Manual testing for now by configuring a self-signed cert for the operator pod and then updating it via cert-manager, which triggers the cert to get re-issued and the secret to get updated. K8s handles updating the files on the pod when the secret updates. Here's some logging from a test run:
```
2021-08-31T20:58:22.783Z	INFO	setup	Received fsnotify event:	{"event": "\"/etc/ssl/solr/client-cert/tls.crt\": REMOVE"}

2021-08-31T20:58:22.783Z	INFO	setup	mTLS cert file updated	{"certPath": "/etc/ssl/solr/client-cert/tls.crt", "certFile": "/etc/ssl/solr/client-cert/..2021_08_31_20_58_22.684994880/tls.crt"}

2021-08-31T20:58:22.783Z	INFO	setup	Configured the custom CA pem for the mTLS transport	{"path": "/etc/ssl/solr/ca-cert/rootSolrCert.pem"}

2021-08-31T20:58:22.783Z	INFO	setup	Updated mTLS Http Client after update to cert	{"certPath": "/etc/ssl/solr/client-cert/tls.crt"}

2021-08-31T20:58:22.783Z	INFO	setup	Re-added watch for symlink to mTLS cert file	{"certPath": "/etc/ssl/solr/client-cert/tls.crt", "certFile": "/etc/ssl/solr/client-cert/..2021_08_31_20_58_22.684994880/tls.crt"}

2021-08-31T20:58:36.153Z	INFO	controllers.SolrCloud	Updating SolrCloud Status	
```

### DIFF
--- a/helm/solr-operator/templates/deployment.yaml
+++ b/helm/solr-operator/templates/deployment.yaml
@@ -64,6 +64,10 @@ spec:
         {{- if .Values.mTLS.insecureSkipVerify }}
         - --tls-skip-verify-server={{ .Values.mTLS.insecureSkipVerify }}
         {{- end }}
+        {{- if .Values.mTLS.watchForUpdates }}
+        - --tls-watch-cert={{ .Values.mTLS.watchForUpdates }}
+        {{- end }}
+
         env:
           - name: POD_NAMESPACE
             valueFrom:
--- a/helm/solr-operator/templates/deployment.yaml
+++ b/helm/solr-operator/templates/deployment.yaml
@@ -64,9 +64,7 @@ spec:
         {{- if .Values.mTLS.insecureSkipVerify }}
         - --tls-skip-verify-server={{ .Values.mTLS.insecureSkipVerify }}
         {{- end }}
-        {{- if .Values.mTLS.watchForUpdates }}
         - --tls-watch-cert={{ .Values.mTLS.watchForUpdates }}
-        {{- end }}
 
         env:
           - name: POD_NAMESPACE
--- a/build/NOTICE-ADDITION
+++ b/build/NOTICE-ADDITION
@@ -1,6 +1,9 @@
 This project uses the hashicorp/golang-lru project, which is MPL 2.0 licensed. The source code can be found at
     https://github.com/hashicorp/golang-lru
 
+This project uses the runtime dependency https://github.com/fsnotify/fsnotify, which is BSD-3 licensed.
+See the below notice, provided by the project.
+
 =================================================
 ==  FSNotify Notice                            ==
 =================================================
--- a/build/NOTICE-ADDITION
+++ b/build/NOTICE-ADDITION
@@ -1,2 +1,5 @@
 This project uses the hashicorp/golang-lru project, which is MPL 2.0 licensed. The source code can be found at
     https://github.com/hashicorp/golang-lru
+
+This project uses the fsnotify/fsnotify library, which is BSD 3-Clause "New" or "Revised" licensed. The source code can be found at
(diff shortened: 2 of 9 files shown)


helm/solr-operator/templates/deployment.yaml
```diff
@@ -64,6 +64,10 @@ spec:
         {{- if .Values.mTLS.insecureSkipVerify }}
         - --tls-skip-verify-server={{ .Values.mTLS.insecureSkipVerify }}
         {{- end }}
+        {{- if .Values.mTLS.watchForUpdates }}
+        - --tls-watch-cert={{ .Values.mTLS.watchForUpdates }}
```
HoustonPutman, line 68: `{{- if .Values.mTLS.watchForUpdates }}` returns false for `false` values, so the only way for the if statement to be passed is if `watchForUpdates` is true. Since the default value in the operator is `true`, there's no way to set `watchForUpdates` to *false* using the helm chart.

I would probably just remove the if statement here.

HoustonPutman (changes requested):
Would be good to add a changelog entry as well.

Awesome work with the fsnotify stuff. It looks like that is a BSD3 licensed library, so I think you need to include the [License text](https://github.com/fsnotify/fsnotify/blob/master/LICENSE) in the [notice-additions file](https://github.com/apache/solr-operator/blob/main/build/NOTICE-ADDITION) that generates the notice file found in the docker image.

helm/solr-operator/templates/deployment.yaml
thelabdude: ha ha! good catch, was just copy-and-pasting senselessly ...

thelabdude: verified by doing:
```
helm template helm/solr-operator --set mTLS.watchForUpdates=false
```
resulting config is:
```
       args:
        - -zk-operator=true
        - --tls-skip-verify-server=true
        - --tls-watch-cert=false
```

thelabdude:
> Would be good to add a changelog entry as well.
> 
> Awesome work with the fsnotify stuff. It looks like that is a BSD3 licensed library, so I think you need to include the [License text](https://github.com/fsnotify/fsnotify/blob/master/LICENSE) in the [notice-additions file](https://github.com/apache/solr-operator/blob/main/build/NOTICE-ADDITION) that generates the notice file found in the docker image.

Change log entry added

Not sure if I did the NOTICE correctly? just followed what was there ...

HoustonPutman (approved):
Beautiful 👍

Issue #322, opened 2021-09-14
Solr pods graceful shutdown 


vladiceanu:
### Context
In Kubernetes, from the moment you issue `kubectl delete pod` command and until the Pod is deleted there are a few steps that happen: 
1. endpoint is being removed from the Endpoints (k8s object);
2. if the pod has a `preStop` hook set, it will run it before `SIGTERM` is invoked;
3. control-plane fires an event to Kube-proxy, CoreDNS, Ingress controller to deregister the Pod's IP address, and no further traffic should be sent to it
**AND, in parallel, after the `preStop` hook finishes,**
app receives `SIGTERM` and, if it's able to process it, it starts graceful shutdown; otherwise - wait for the `terminationGracePeriodSeconds` (default 30s) period to pass, and then `SIGKILL` is fired;
4. pod is deleted;

### Graceful shutdown
In order to achieve a graceful shutdown, we must satisfy the following condition(s): no traffic is sent to the non-existent IP (pod already deleted). That should be done in Step 3 described above, but, since those components (kube-proxy, coredns, ingress controller) might be busy with something else, **there is no guarantee that the IP will be removed from their state before the Pod is gone**. How long would it take? It depends; some of them might take less than a second, the others a bit longer.  

**Race condition**
As mentioned, deregistration of the IP from kube-proxy, CoreDNS, ingress controller, and `SIGTERM` sent to the APP happens in parallel, which can cause a few race conditions, one of them is: what if pod is deleted before the IP is deregistered? That could be a problem, since traffic might be sent to a non-existent IP.

### Issue statement
Graceful shutdown of SolrCloud. Currently, we use `preStop` hook where we run `solr stop -p 8983` (which kernel behind the scenes sends `SIGQUIT` to the process) which stop solr instances on port 8983 that run in the background. 

But, as we already know, `preStop` hook (step 2) is executed before kube-proxy, coredns, ingress controller received the event to deregister the IP address from their local state (step 3) and it will stop the Solr instance before deregistering its IP, , thus, traffic will be sent to a non-existent IP.

A few ways to handle that:
- custom input for `lifecycle.preStop` parameter that can be passed by the user and will overwrite the default `solr stop -p 8983`; 
- additional command before `solr stop -p 8983` in `preStop` hook that can be passed by the user, i.e. `lifecycle.preStop.cmd: sleep 30`, when merged, we'll get something like `sleep 30 && solr stop -p 8983`;
[...]

PR #324, opened 2021-09-20
Support custom Lifecycle for Solr & PrometheusExporter containers

vladiceanu:
Resolves https://github.com/apache/solr-operator/issues/322 

**Describe your changes**
This PR adds support for custom Lifecycle input for Solr main container. A user should be able to set `lifecycle.preStop` and `lifecycle.postStart` hooks through SolrCloud CR for Solr main container.

This PR also includes Helm chart changes to support lifecycle inputs.

Tested locally on minikube.


controllers/util/solr_util.go
```diff
@@ -327,6 +328,13 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 		}
 	}
 
+	// Default preStop hook
+	preStop := &corev1.Handler{
```
HoustonPutman, line 332: Is there a reason you split this out here?

HoustonPutman (changes requested):
This looks good to me! But I would like to see a test added 🙂 

Great contribution 🎉

controllers/util/solr_util.go
vladiceanu: Just for better readability, I guess.

vladiceanu: > This looks good to me! But I would like to see a test added 🙂

I was checking around and found that we already have a few tests on Lifecycle, i.e. https://github.com/apache/solr-operator/blob/main/controllers/solrcloud_controller_externaldns_test.go#L117, https://github.com/apache/solr-operator/blob/main/controllers/solrcloud_controller_test.go#L300, https://github.com/apache/solr-operator/blob/main/controllers/solrcloud_controller_test.go#L396, and other places.

Are there any other kinds of tests that you think it would make sense to add? Perhaps just a simple custom input, not based on variables?

HoustonPutman: Yeah, if you look at https://github.com/apache/solr-operator/blob/main/controllers/solrcloud_controller_test.go#L174, you'll see we test basically all of the custom options there. Add the custom lifecycle, and test that the podSpec actually uses it.

vladiceanu: Make sense. I've cleaned up the tests of `preStop` hook that were based on ENV variable `SOLR_PORT` because we're passing different Lifecycle values and that won't work; in any case, we already have a test for ENV variables. 
Also, I have added the Lifecycle tests for SolrCloud and PromExporter STS/Deployment. 

Please have a look and let me know if I miss anything.

HoustonPutman: The new tests look great, but why did you remove the old tests from the tests that don't specify a custom lifecycle?

We test those because the ports change between the tests, and we want to make sure the correct port is always used no matter the solrAddressibility options specified.

It looks like we are keeping the same default lifecycle, unless I'm missing something. So those tests wouldn't have to change, I don't think.

vladiceanu: oh, yes, I thought it doesn't make sense, but now I see that the point of these tests is to ensure the correctness of the used port. I have added them back.

vladiceanu: thank you for all the help, Houston.

HoustonPutman: No worries, thanks for the great and thorough PR!

HoustonPutman (approved):
Great work!

Issue #331, opened 2021-09-30
Support JWT authentication


janhoy:
The Operator already has support for configuring Solr with BasicAuth, and the Operator is given BasicAuth credentials to be able to access `/solr/admin/info/system`, `/solr/admin/collections` and `/solr/admin/{backup,restore}`.

The operator should also work when [JWTAuthPlugin](https://solr.apache.org/guide/8_10/jwt-authentication-plugin.html) is used for the cluster. The Operator will then need to obtain and use a JWT token as an `Authorization: Bearer xxxx` header for all requests to Solr endpoints, analogous to how it uses `Authorization: Basic xxxx` today.

What I propose is
* Operator can be configured to setup `security.json` for JWT
* Operator can be configured to obtain a JWT token from an OIDC server to talk to Solr

The security.json is not that different from the BasicAuth one, something like:

```json
{
  "authentication": {
    "blockUnknown": false,
    "class": "solr.JWTAuthPlugin",
    "redirectUris": "https://my.solr.server:8983/solr/,https://my.other.solr.server:8983/solr/",
    "rolesClaim": "roles",
    "issuers": [
      {
        "wellKnownUrl": "https://idp.example.com/.well-known/openid-configuration",
        "clientId": "<MY_CLIENT_ID>",
      }
    ]
  },
  "authorization": {
    "class": "solr.ExternalRoleRuleBasedAuthorizationPlugin",
    "permissions": [ ... ]
  }
}
```

The permissions will mostly be the same, and the mapping from users to their roles will happen in the OIDC server, so we'll not care about usernames in security.json, just roles. To generate this, we need some more config values to operator:

```yaml
spec:
  ...
  solrSecurity:
    authenticationType: JWT
    jwt:
      wellKnownUrl: <url>
      solrClientId: <solr client-id as registered with OIDC>
      rolesClaim: <jwt claim key where role name is stored>
      oper-role: k8s
      admin-role: admin
      operClientId: <operator client-id as registered with OIDC>
      operClientSecretName: <name of k8s secret where operator's client secret for OIDC is stored>
```

The user will ahead of time register Solr and SolrOperator with OIDC server to obtain client-ID and secret. The Operator will generate and provision `security.json` and connect to OIDC's token endpoint to obtain a JWT token for Solr.

[...]

PR #334, opened 2021-10-01
Refactor security related code into a separate solr_security_util.go vs. sprinkled throughout the codebase.

thelabdude:
There's probably more we can do with the security code, but I'll tackle additional changes when I work on #331 as most of the remaining changes to the security code would be related to supporting a different authn scheme (OIDC) vs. just basic auth as the code is doing now.

### DIFF
--- a/controllers/util/solr_security_util.go
+++ b/controllers/util/solr_security_util.go
@@ -1,0 +1,434 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


controllers/util/solr_security_util.go
```diff
@@ -0,0 +1,434 @@
...
+)
+
+type SecurityConfig struct {
+	BasicAuthSecret *corev1.Secret
+	SecurityJson    string
+}
+
+// Given a SolrCloud instance and an API service client, produce a SecurityConfig needed to enable Solr security
+func ReconcileSecurityConfig(client *client.Client, instance *solr.SolrCloud) (*SecurityConfig, error) {
+	ctx := context.TODO()
```
HoustonPutman, line 53: can you pass the context from the controller, now that a reconcile is called with a given context?

controllers/util/solr_util.go
```diff
@@ -1010,21 +992,21 @@ func generateZKInteractionInitContainer(solrCloud *solr.SolrCloud, solrCloudStat
 
 	cmd := ""
 
+	if hasChroot {
+		cmd += "solr zk ls ${ZK_CHROOT} -z ${ZK_SERVER} || solr zk mkroot ${ZK_CHROOT} -z ${ZK_SERVER}; "
+	}
+
 	if solrCloud.Spec.SolrTLS != nil {
 		cmd = setUrlSchemeClusterPropCmd()
```
HoustonPutman, line 1000: I think you want this to be a `+=`, if you are moving the `hasChroot` logic above here.

HoustonPutman (commented):
Love all PRs that make solr_util and solrcloud_controller smaller!

Just a few things I noticed.

PR #343, opened 2021-10-13
Use global.imagePullSecrets at solr-operator helm chart #338

### LINKED ISSUE #338: Use global.imagePullSecrets at solr-operator helm chart
In some scenarios, having to use pull secrets is mandatory. Even if using a private repo, or pulling from the public repos, but hitting the dockerhub anonymous pull limit (https://www.docker.com/increase-rate-limits).

The idea would be to use the `global.imagePullSecrets`, as in the main solr(cloud) helm, or in the zookeeper-operator.

j-coll:
Same PR as #339, but against main branch


HoustonPutman (changes requested):
Thanks for the contribution, this would indeed be very helpful for people!

Before merging would you mind adding a Changelog entry here:
https://github.com/apache/solr-operator/blob/main/helm/solr-operator/Chart.yaml#L56

And adding an entry to the helm chart documentation as well?
https://github.com/apache/solr-operator/blob/main/helm/solr-operator/README.md#running-the-solr-operator

HoustonPutman:
Well if you want to share imagePullSecrets here, you can merely use the same serviceAccount that is created by the Zookeeper Operator with the Solr Operator. That way the imagePullSecrets will be available to both through the same serviceAccount.

```yaml
serviceAccount:
  create: false
  name: operator-account

zookeeper-operator:
  install: true
  serviceAccount:
     create: true
     name: operator-account
     imagePullSecrets:
        - list
        - of
        - strings
```

Using the above helm values, the zookeeper-operator helm chart will create the serviceAccount with the listed imagePullSecrets. Then the Solr Operator will merely use the same one, named `operator-account`.

No need to actually use the global imagePullSecrets.

piotrrybicki:
Hello there. @j-coll are you willing to do proper changes in this PR?
I'd love to have ability to provide imagePullSecrets in globals. I'm not using ZK operator.

I can also create my own PR about this issue.

HoustonPutman:
@piotrrybicki Please open your own PR. We'd love to support as many globals as possible, to make setup easy for everyone!

As for my comment above it would be great if the helm code could detect a list of strings and convert it to a list of objects where necessary.

janhoy:
@j-coll Are you still planning to land this?

HoustonPutman:
Closing in favor of #832 . Thanks for the work here, sorry we weren't able to get it through before.

PR #346, opened 2021-10-13
Ability to get a JWT token from an OIDC provider to make calls to Solr secured by the JWTAuthPlugin

### LINKED ISSUE #331: Support JWT authentication

thelabdude:
Fixes #331 

Manually tested with Keycloak.

I went with calling this AuthenticationType `Oidc` even though technically, we're only using the `client_credentials` grant type in OAuth2 (see: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4); there's no redirecting or any of that fancy OIDC stuff. I like `Oidc` because that corresponds better with the `JWTAuthPlugin` used by Solr and our config takes a `wellKnownUrl` which is an OIDC thing.

Example SolrCloud CRD YAML:
```
spec:
  ...
  solrSecurity:
    authenticationType: Oidc
    probesRequireAuth: false
    bootstrapSecurityJson:
      name: security-jwt-keycloak
      key: security.json
    oidc:
      clientCredentialsSecret: keycloak-gatekeeper
      wellKnownUrl: http://192.168.86.21:8080/auth/realms/local
```

security.json for Keycloak:
```
{
  "authentication": {
    "blockUnknown": false,
    "class": "solr.JWTAuthPlugin",
    "wellKnownUrl": "http://192.168.86.21:8080/auth/realms/local/.well-known/openid-configuration",
    "clientId": "gatekeeper",
    "redirectUris": "http://localhost:8983/solr/",
    "rolesClaim": "groups"
  },
  "authorization": {
    "class": "solr.ExternalRoleRuleBasedAuthorizationPlugin",
    "permissions": [ ... ]
  }
}
```

Todo:
- [x] Unit tests
- [x] Oidc support for the backup CRD controller
- [x] Manual test with another OIDC provider such as Okta
- [x] Manual test to verify the cached OAuth2 TokenSource can refresh the JWT when it expires
- [x] Enhance documentation
- [x] OAuth2 TokenSource cache invalidation, esp. after SolrCloud object is deleted, or just maybe a TTL + LRU kind of strategy

The biggest issue I see with adding this feature is the Prometheus exporter and the bin/solr script don't support the JWTAuthPlugin (see: https://issues.apache.org/jira/browse/SOLR-12666) so users would have to open up access to anonymous requests for those endpoints. So I wonder if the better approach is to improve Solr to support multiple authentication schemes based on the provided Authorization header either `Bearer` or `Basic`


thelabdude:
Now this PR only has OIDC related changes in it. However, it's unlikely I'll merge this into main anytime soon given the limitations with JWT / OIDC on the Solr side (see previous comments); at least not until SolrJ supports OIDC (so that the prometheus exporter works).

janhoy:
Closing this. Can revive it in a new PR some time later if necessary

ppazos:
@thelabdude following this thread, related to my current issue, do you know if a JWT, to be used in the Solr API, can be generated without a human user login into a web form? Because we have the case of system to system auth, leaving the end user GUI open for searching. Thanks!

janhoy:
You can use a MultiAuthPlugin in Solr and use JWT for humans and BasicAuth for Operator and Prom Exporter.

ppazos:
@janhoy I want to use JWT for app to solr, not for humans, my search app is open for users but I don't want other apps to talk to solr, is that possible? I like tokens to expire and be renewed, instead of using basic auth. Having app level auth would be useful in this scenario. Thanks.

janhoy:
To use Multi with the operator you'll need to craft the security.json yourself (e.g. through a Helm template), put it in a secret, and then tell Solr Operator to use that secret. I don't think you'll get much help from the operator itself.
For further discussion, please open a new email thread in the users mailing list.

PR #349, opened 2021-10-14
Allow for managed restarts across multiple SolrCloud resources.

### LINKED ISSUE #348: Issue with rolling restart with multi SolrCloud clusters
In my testing if I had a single Solr collection with a SolrCloud object per shard, the managed rolling restart would not restart the pod with the overseer.

HoustonPutman:
Fixes #348 

This is a change to improve the logic of the managed upgrades, so that the solr operator does not assume that a cloud is completely managed by one SolrCloud resource.

Instead of making sure that `# of desired pods` = `# of live nodes` before upgrading the overseer, the operator now just makes sure that all pod names that it manages are accounted for in the cluster's live nodes. This allows for running multiple SolrCloud resources that connect together, while still using the managed restart functionality.

Note this is a fairly small change, but required a lot of refactoring in the tests. That is mostly noise.


HoustonPutman:
@joel-bernstein, could you build this locally and try it out with your setup? The unit tests are pretty comprehensive, but it would be good to have some live testing as well!

PR #357, opened 2021-10-27
Remove Persistence option for SolrBackup.

### LINKED ISSUE #347: Remove SolrBackup Persistence option
With the support for `GCS` and `S3` backups (#301, #328), we should remove the Persistence options for the SolrBackup.

It's really just a legacy option to help people store things in S3, before the S3Repository was made available. It's janky and doesn't work with the new incremental backups.

I think it would be best to just remove the option all-together.

HoustonPutman:
Resolves #347 

TODO list:

- [x] Upgrade docs & ChangeLog entry
- [x] Local Testing


config/crd/bases/solr.apache.org_solrbackups.yaml
```diff
@@ -71,7 +71,7 @@ spec:
                 description: The location to store the backup in the specified backup repository.
                 type: string
               persistence:
-                description: Persistence is the specification on how to persist the backup data.
+                description: 'Persistence is the specification on how to persist the backup data. This feature has been removed as of v0.5.0. Any options specified here will not be used. TODO: Remove this field entirely in v0.6.0'
```
gerlowskija, line 74: [0] It's a little odd to see a TODO comment in what (I think?) is user-facing documentation.  Should this be in a `#`-based YAML comment instead?

config/crd/bases/solr.apache.org_solrbackups.yaml
```diff
@@ -1106,7 +1106,7 @@ spec:
                 description: Whether the backup has finished
                 type: boolean
               persistenceStatus:
-                description: Whether the backups are in progress of being persisted
+                description: 'Whether the backups are in progress of being persisted. This feature has been removed as of v0.5.0. TODO: Remove this field entirely in v0.6.0'
```
gerlowskija, line 1109: [0] ditto re: maybe the TODO should be a comment?

gerlowskija (approved):
I like the idea of removing this functionality - especially since the recent addition of s3 backup support supplants the original intended use case for "persistence".

The PR looks good to me.  I left a niggling comment or two, but nothing that strictly needs addressed IMO.

config/crd/bases/solr.apache.org_solrbackups.yaml
HoustonPutman: Fair enough on removing the TODO, because we might want to leave in the field anyways (for yaml back-compat).

These files are auto-generated by `make manifests` after you edit anything in the `api/` package. So we can't really add yaml comments here. These descriptions are taken from the comments before the property in the struct.

PR #358, opened 2021-10-27
Specify individual backupRepo availability in SolrCloud Status

### LINKED ISSUE #326: Break up backupIsReady in SolrCloud status by backupRepository
There is also a much easier way to calculate this.

When we generate a Solr pod spec in the StatefulSet, that pod spec should "be ready" for all backupRepos currently configured. So if we set an annotation on the pod for each repo stating it's most recent version, then we should be able to easily get the annotations on each pod to see the backupRepos that are able to be used on each pod. Much better than scanning for various volumes, or whatever else that specific BackupRepository implementation might require.

HoustonPutman:
Resolves #326 

This uses a map of the BackupRepositories names in the SolrCloud status, mapping the name of the repo to a boolean of whether that repo is available across all nodes.

TODO:

- [x] Add changelog entry
- [x] Manual testing


config/crd/bases/solr.apache.org_solrbackups.yaml
```diff
@@ -1053,9 +1053,12 @@ spec:
                 type: object
               repositoryName:
                 description: The name of the repository to use for the backup.  Defaults to "legacy_local_repository" if not specified (the auto-configured repository for legacy singleton volumes).
+                pattern: '[a-zA-Z0-9]([-_a-zA-Z0-9]*[a-zA-Z0-9])?'
                 type: string
               solrCloud:
                 description: A reference to the SolrCloud to create a backup for
+                minLength: 63
```
gerlowskija, line 1060: [-1] I imagine you put this in for testing at some point or something?

helm/solr-operator/crds/crds.yaml
```diff
@@ -1053,9 +1053,12 @@ spec:
                 type: object
               repositoryName:
                 description: The name of the repository to use for the backup.  Defaults to "legacy_local_repository" if not specified (the auto-configured repository for legacy singleton volumes).
+                pattern: '[a-zA-Z0-9]([-_a-zA-Z0-9]*[a-zA-Z0-9])?'
                 type: string
               solrCloud:
                 description: A reference to the SolrCloud to create a backup for
+                minLength: 63
```
gerlowskija, line 1060: [-1/Q] I commented above on this value in `solr.apache.org_solrbackups.yaml`, suspecting it was something you'd set for testing but hadn't removed yet.  If that ends up being the case, I figured I'd comment here as a reminder to make sure that your change gets propagated to this file as well.

helm/solr-operator/crds/crds.yaml
```diff
@@ -1053,9 +1053,12 @@ spec:
...
+                pattern: '[a-z0-9]([-a-z0-9]*[a-z0-9])?'
```
gerlowskija, line 1061: [-1/Q] Ditto to my reminder comment above, but this time regarding the lower-case requirement on this string field that you may or may not have intended.

gerlowskija (commented):
This looks great!

There were a few things I noticed in the crd spec that seemed like they might've been oversights (esp inconsistency around case-sensitivity and min-length).  But otherwise it looks "ready to go" to me.

It's got my 👍  as long as you lmk whether those crd spec things were intended or not.

config/crd/bases/solr.apache.org_solrbackups.yaml
HoustonPutman: nah, I actually meant maxLength here (because kubernetes resource names cannot go beyond 63 characters: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)

Good find!

api/v1beta1/solrcloud_types.go
```diff
@@ -385,6 +385,8 @@ type SolrBackupRestoreOptions struct {
 type SolrBackupRepository struct {
 	// A name used to identify this local storage profile.  Values should follow RFC-1123.  (See here for more details:
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)
+	//
+	// +kubebuilder:validation:Pattern:=[a-zA-Z0-9]([-_a-zA-Z0-9]*[a-zA-Z0-9])?
```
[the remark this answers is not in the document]
HoustonPutman: So this is just a name for the backup repository, it is pretty independent from the name of the SolrCloud (which has limitations given by Kubernetes, which I linked below).
I used the same regex, but changed it to allow capital letters and underscores, since we don't need to be as strict with the names of the repositories (as far as I know).

Is there any limitation in Solr for the name of backup repositories? If so we should definitely change this to match.

helm/solr-operator/crds/crds.yaml
HoustonPutman: So yeah, just to be very clear, I'm just following the guidelines for [kubernetes resource names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). This field is a reference to SolrCloud resources, so there is no way that giving a string that has a length > 63, or doesn't match that regex, is going to be successful (because we will never be able to find a SolrCloud in kubernetes with the given name).

HoustonPutman:
Will test (and fix if need be) and merge tomorrow.

PR #359, opened 2021-10-28
Add ability to schedule recurring incremental backups

### LINKED ISSUE #303: Add scheduling support for backups
Currently, each ‘solrbackup’ resource triggers a one-time backup of the specified collections.  This is great for many usecases that only need backups on an ad-hoc basis, but it falls short on SaaS or other hosted usecases that often take backups on a regular schedule as a basic precaution.

Operator users should be able to configure a schedule (either in the existing ‘solrbackup’ CRD, or in a new ‘solrbackupschedule’ resource type) so that backups can be set up to recur without further intervention.  Ideally some sort of “maxBackupsRetained” setting would also be added to garbage-collect old backups that are no longer needed I guess?

HoustonPutman:
Resolves #303 

Still needs a bit of love, but the bulk of the work should be there.

TODO:

- [x] Documentation & Changelog entry
- [x] Cleaning up logic around restarts
- [x] Rebase after other backup PRs are merged
- [x] Manual testing

### DIFF
--- a/api/v1beta1/solrbackup_types.go
+++ b/api/v1beta1/solrbackup_types.go
@@ -94,6 +94,16 @@ type BackupRecurrence struct {
 	// +kubebuilder:validation:Minimum:=1
 	// +optional
 	MaxSaved int `json:"maxSaved,omitempty"`
+
+	// Disable the recurring backups. Note this will not affect any currently-running backup.
+	//
+	// +kubebuilder:default:=false
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
+}
+
+func (recurrence *BackupRecurrence) IsEnabled() bool {
+	return recurrence != nil && !recurrence.Disabled
 }
 
 // PersistenceSource defines the location and method of persisting the backup data.
--- a/config/crd/bases/solr.apache.org_solrbackups.yaml
+++ b/config/crd/bases/solr.apache.org_solrbackups.yaml
@@ -1063,6 +1063,10 @@ spec:
               recurrence:
                 description: "Set this backup to be taken recurrently, with options for scheduling and storage. \n NOTE: This is only supported for Solr Clouds version 8.9+, as it uses the incremental backup API."
                 properties:
+                  disabled:
+                    default: false
+                    description: Disable the recurring backups. Note this will not affect any currently-running backup.
+                    type: boolean
                   maxSaved:
                     default: 5
                     description: Define the number of backup points to save for this backup at any given time. The oldest backups will be deleted if too many exist when a backup is taken. If not provided, this defaults to 5.
--- a/controllers/solrbackup_controller.go
+++ b/controllers/solrbackup_controller.go
@@ -96,8 +96,8 @@ func (r *SolrBackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	// Check if we should start the next backup
 	if backup.Status.NextScheduledTime != nil {
-		// If the backup no longer has a recurrence specified, remove the next scheduled time
-		if backup.Spec.Recurrence == nil {
+		// If the backup no longer enabled, remove the next scheduled time
+		if !backup.Spec.Recurrence.IsEnabled() {
 			backup.Status.NextScheduledTime = nil
 			backupNeedsToWait = false
 		} else if backup.Status.NextScheduledTime.UTC().Before(time.Now().UTC()) {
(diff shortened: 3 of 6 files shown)


api/v1beta1/solrbackup_types.go
```diff
@@ -64,6 +71,29 @@ func (spec *SolrBackupSpec) withDefaults(backupName string) (changed bool) {
...
+	Schedule string `json:"schedule"`
+
+	// Define the number of backup points to save for this backup at any given time.
+	// The oldest backups will be deleted if too many exist when a backup is taken.
+	// If not provided, this defaults to 10.
+	//
+	// +kubebuilder:default:=10
+	// +kubebuilder:validation:Minimum:=1
+	// +optional
+	MaxSaved int `json:"maxSaved,omitempty"`
```
gerlowskija, line 94: [0] This is mostly subjective preference, but I'd vote for a smaller default value here.  For collections with a lot of churn, 10 backups might be a lot of data to store by default.  Would you be open to a lower default like 3 or 5?

gerlowskija (changes requested):
This looks awesome!

All of the code itself looks good (aside from tests which I imagine you just haven't gotten to yet).  I do have two suggestions/question about the feature overall though: 

1. Confusingly, incremental backup 'deletion' doesn't really do a full delete: instead it removes only the metadata associated with the backup, potentially leaving index-files behind.  Any backed up index-files that are no longer used must then be deleted by a separate garbage-collection process called "[purging](https://solr.apache.org/guide/8_10/collection-management.html#deletebackup-parameters)."
IMO we should handle this purging process for users: either by triggering a "purge" on every backup, or by exposing a `purgeEvery` setting alongside `maxSaved` to allow a less aggressive approach. 
2. I can imagine a lot of users might use recurrence but want some way to temporarily disable it while they do a perf test, upgrade, bulk data loading, etc.  What do you think about putting an "enabled" flag in `BackupRecurrence` to let users temporarily disable things without totally deleting the solrbackup?

api/v1beta1/solrbackup_types.go
HoustonPutman: Very open. I literally just chose a number here. I am very happy to differ to whatever you think it should be. Give the call and it is done.

HoustonPutman:
1. I'm content with a `purgeEvery` option in Solr. I'm not sure I understand why you would want un-garbage collected index files laying about, but that's definitely a different issue. We can definitely go ahead and purge after every backup until the `purgeEvery` option is added. (Oh now I realize you mean a `purgeEvery` option in the CRD. I think we can assume that would be equivalent to `saveEvery`.)
I did talk with @anshumg about a ticket that saw some issues with purging that does make me a bit wary to do it always. https://issues.apache.org/jira/browse/SOLR-15706
We should make sure that's not an issue before going forward with that.
2. That's a good idea. I do think I want to have it "enabled" by default. Which is easy to do with those kubebuilder annotations.

gerlowskija:
> now I realize you mean a `purgeEvery` option in the CRD. I think we can assume that would be equivalent to saveEvery.

Yep, that's what I was suggesting - an option in the CRD to expose the option that's already available in Solr.  That SOLR-15703 you pointed to looks very serious 😬 .  I'll take a look and hopefully get the Solr concern out of the way so we can consider exposure in the operator.

When you mention "saveEvery" I assume you mean maxSaved?  Or is there some other "saveEvery" param I missed?

HoustonPutman:
> When you mention "saveEvery" I assume you mean maxSaved? Or is there some other "saveEvery" param I missed?

I mis-typed. I meant `maxSaved` 🙂

HoustonPutman:
@gerlowskija, given that https://issues.apache.org/jira/browse/SOLR-15706 is not released yet, maybe we just wait to add in the `purge` call until the next Solr Operator release.

api/v1beta1/solrbackup_types.go
HoustonPutman: I set the default to 5

HoustonPutman:
@gerlowskija almost forgot, but I added the `disabled` flag and it works really well!

killdash9:
> @gerlowskija, given that https://issues.apache.org/jira/browse/SOLR-15706 is not released yet, maybe we just wait to add in the `purge` call until the next Solr Operator release.

That issue has been fixed.  Should we add purgeUnused to the solr operator's backup functionality?

HoustonPutman:
Absolutely, you should create a new issue for that @killdash9 !

PR #360, opened 2021-10-29
Enable Solr Operator metrics when using the helm chart.

### LINKED ISSUE #307: Expose operator metrics
Users may want to monitor the operator health. See https://book.kubebuilder.io/reference/metrics.html for some docs about publishing Prometheus metrics.

No idea what metrics are most useful. Some counters for various operations and errors. Perhaps a gauge for number of SolrClouds deployed?

HoustonPutman:
Resolves #307 

This does not use the protected metrics-endpoint described in the [kubebuilder docs](https://kubebuilder.io/reference/metrics.html#protecting-the-metrics). This is also just using the default operator metrics. We can add additional metrics in future PRs.

The metrics have been enabled since #321 was implemented, but not when using the helm chart. This fixes that.

### DIFF
--- a/helm/solr-operator/Chart.yaml
+++ b/helm/solr-operator/Chart.yaml
@@ -148,6 +148,13 @@ annotations:
           url: https://github.com/apache/solr-operator/issues/355
         - name: Github PR
           url: https://github.com/apache/solr-operator/pull/356
+    - kind: added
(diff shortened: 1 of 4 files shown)


janhoy (commented):
LGTM. Not tested

HoustonPutman:
Tested manually and was able to query the operator metrics through `kubectl port-forward`.

Issue #301, opened 2021-08-05
Add support for GCS storage to 'solrbackup'


gerlowskija:
Currently the 'solrbackup' resource assumes that users want backups stored "locally" (i.e. stored on a PV or mounted drive using Solr's LocalFileSystemRepository).  These local backups can then optionally be "persisted" - which involves compressing them and shipping them to a different PV or S3 bucket.

But no support exists for using other backup destinations that Solr supports natively, such as GCS (as of 8.9).

We should add this support.  Users can configure their GCS-backup settings under solrcloud's `backupRestoreOptions` object, leaving the `solrbackup` object relatively untouched (except that any "persistence" section on 'solrbackup' would now be ignored, as we can only easily compress files that are stored locally).

gerlowskija:
Some questions this raises:

- should backup mechanisms be mutually exclusive in a given 'solrcloud' definition, or should we allow users to configure multiple and choose between them in each 'solrbackup'.
- What's the best way to surface an error if a user attempts to configure GCS backups but is using a solr version that doesn't support that.
- should the operator attempt to create any missing GCS paths/buckets using the provided credentials, or require everything to be created up front by users.

gerlowskija:
I've attached a rough PR that shows how this could be done.  Below are an example 'solrcloud' and 'solrbackup' that use the proposed functionality:

**SolrCloud**
```
apiVersion: solr.apache.org/v1beta1
kind: SolrCloud
metadata:
  name: jasons_cluster
spec:
  dataStorage:
    persistent:
      reclaimPolicy: Delete
      pvcTemplate:
        spec:
          resources:
            requests:
              storage: "5Gi"
    backupRestoreOptions:
      gcsStorage:
        bucket: "solr-log-test"
        gcsCredentialSecret: "my-gcs-secret"
        baseLocation: "logs"
    ...
```

**SolrBackup**
```
apiVersion: solr.apache.org/v1beta1
kind: SolrBackup
metadata:
  name: gcs_techproducts_backup
  namespace: default
spec:
  solrCloud: jasons_cluster
  collections:
    - techproducts
```
(Note that there's no new configuration in 'solrbackup', just the removal of the 'persistence' section for gcs-backups.)

I'm not wedded to these syntaxes by any means - just wanted to get some examples up here as a concrete starting point for discussion.

HoustonPutman:
So I just want to make sure I understand correctly.

The only thing the operator should do for these "native" backup options, is to call the Solr API right?
[...]

PR #368, opened 2021-11-09
Change managed backup repo to volume repo.

HoustonPutman:
Related to #347 and #301 

Since the persistence option has been removed, there is no more "management" done for Managed Backup Repositories.
Therefore it makes sense to rename "Managed" backups to "Volume" backups, since it is merely backing up to Kubernetes Volumes.

I'd appreciate thoughts @gerlowskija . I'd like to get this in before creating the `v0.5.0` RC1, so we don't have to deal with backwards compatibility when making the change.


gerlowskija (approved):
LGTM - you're right that the name "managed" doesn't make much sense anymore.  And even when it did, there was some inconsistency with the term "local".

Happy to see this get in!

PR #371, opened 2021-11-09
Remove mirrors from release wizard

HoustonPutman:
https://blogs.apache.org/foundation/entry/apache-software-foundation-moves-to

So there are no more mirrors to check. So I am removing the name "mirrors" from the release wizard, as well as removing the script that checks the mirrors.


HoustonPutman:
Copied some logic from here: https://github.com/apache/solr/pull/348

PR #394, opened 2022-01-14
Make GcsRepository.gcsCredentialSecret optional

gerlowskija:
There are some circumstances where GCS access can be managed without
explicitly providing a service account key (e.g. when GKE's "Workload
Identity" feature is used to provide auth metadata).

This commit drops the hard requirement on the `gcsCredentialSecret`
property with this in mind.


HoustonPutman (changes requested):
This looks great!

2 Things:

- Can you add a changelog entry in `helm/solr-operator/Chart.yaml`?
- What version of golang are you using locally? The tests are failing because of the new line in each `generated.deepcopy.go`: `//go:build !ignore_autogenerated`. It seems like you are using a newer version of golang than the operator uses. For this PR do you mind just removing that line in each of the files (`api/v1beta1/zz_generated.deepcopy.go` and `controllers/zk_api/zz_generated.deepcopy.go`). We can upgrade the golang version in another issue/PR.

gerlowskija:
Hmm, looks like I've got `go version go1.17.5 darwin/amd64` locally.

Thanks for the review and the heads up on the build failure - it's been on my list but I hadn't circled back yet.  Gonna address those review comments now.

HoustonPutman:
We could go ahead and upgrade to golang 1.17 as well (though it should be a separate ticket). I was waiting for 1.18, but it will probably be a while until that is good to use in production.

HoustonPutman (approved):
Looks good to me!

Issue #393, opened 2022-01-14
GKE Autopilot constant re-deployment of pods


hrvolapeter:
I'm runnig solr operator on GKE Autopilot clust with example from [here](https://solr.apache.org/operator/articles/explore-v030-gke.html) - the example requires to round-up resource requests to values supported by autopilot. Hovewer, operator keeps rededeploying deployments due to change in resource requsts, but the 'to' and 'from' valus are exactly the same. This bug is best observable with prometheus exporthes which is redeployed couple times per second, the same is happening with cluster but less often - seem like every couple minutes.

Logs, seem like relevant code is [this](https://github.com/apache/solr-operator/blob/94b62dc967ac4c6e1c8f16acf45d43298ca103d3/controllers/util/common.go#L545) - note the to and from clauses
```
controller-runtime.manager.controller.solrprometheusexporter Update required because field changed {"reconciler group": "solr.apache.org", "reconciler kind": "SolrPrometheusExporter", "name": "explore-prom-exporter", "namespace": "default", "deployment": "explore-prom-exporter-solr-metrics", "kind": "deployment", "field": "Spec.Template.Spec.Containers[0].Resources", "from": {"limits":{"cpu":"250m","ephemeral-storage":"1Gi","memory":"512Mi"},"requests":{"cpu":"250m","ephemeral-storage":"1Gi","memory":"512Mi"}}, "to": {"limits":{"cpu":"250m","ephemeral-storage":"1Gi","memory":"512Mi"},"requests":{"cpu":"250m","ephemeral-storage":"1Gi","memory":"512Mi"}}}
```
```
Error
2022-01-13T21:46:02.884596420Z2022-01-13T21:46:02.884Z INFO controller-runtime.manager.controller.solrcloud Update required because field changed {"reconciler group": "solr.apache.org", "reconciler kind": "SolrCloud", "name": "explore", "namespace": "default", "zookeeperCluster": "explore-solrcloud-zookeeper", "kind": "zookeeperCluster", "field": "Spec.Pod.Resources", "from": {"limits":{"cpu":"250m","ephemeral-storage":"512Mi","memory":"500Mi"},"requests":{"cpu":"250m","ephemeral-storage":"512Mi","memory":"500Mi"}}, "to": {"limits":{"cpu":"250m","ephemeral-storage":"512Mi","memory":"500Mi"},"requests":{"cpu":"250m","ephemeral-storage":"512Mi","memory":"500Mi"}}}
```

[...]

PR #398, opened 2022-01-19
Fix resource checking, to not overwrite values that should be equal.

HoustonPutman:
Part of #393

As it turns out, `reflect.DeepEquals()` does not actually work correctly for `Quantity` values. This is because, while the actual quantities themselves could be the same, if they have a different `Format`, then `DeepEquals()` will return `false`. So instead we need to use the `Quantity.Equals()` method, which introduces a lot more logic around the seemingly simple check.

I've introduced unit tests around this new logic. (and hopefully we can start utilizing more unit tests around `common.go`, which is a behemoth)

One problem is that we currently treat an empty resource list and a `nil` resource list differently. We might want to relax this to make sure we are overwriting as little things as possible, when it doesn't really matter.


HoustonPutman:
Ok now nil and empty request lists are treated the same. Will have to do some testing in case this is a problem.

PR #409, opened 2022-02-21
Fix solr prometheus exporter Doc

limingnihao:
Fix solr prometheus exporter Doc. It should be `passwordKey `.


HoustonPutman (approved):

HoustonPutman:
Thanks for the fix @limingnihao !

PR #411, opened 2022-03-03
Add operator version & ArgoCD annotations to CRDs

### LINKED ISSUE #410: Support ArgoCD for the CRDs
When deploying Solr-Operator in [ArgoCD](https://argo-cd.readthedocs.io/en/stable/) using its Helm chart, you may get this error message:
> reason: CustomResourceDefinition.apiextensions.k8s.io "solrclouds.solr.apache.org" is invalid: metadata.annotations: Too long: must have at most 262144 bytes

This is since ArgoCD uses `kubectl apply` by default, which then overflows some buffer. Explanation here: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes

The same does not happen during normal `helm install` since Helm uses another method of installing CRDs.

This is not a bug in solr-operator, but there seems to be a very simple workaround, by adding this annotation to all our CRDs:

```yaml
metadata:
  annotations:
    argocd.argoproj.io/sync-options: Replace=true
```

This will tell Argo to use `kubectl replace` instead of apply, circumventing this limitation.

janhoy:
Add an annotation to our CRDs that will make ArgoCD replace instead of apply.

I have edited the three base CRDs, and also the dependency ZK CRD, but that is fragile, since the ZK crd may be overwritten on next upgrade...

### DIFF
--- a/hack/config/add_crds_annotations.sh
+++ b/hack/config/add_crds_annotations.sh
@@ -30,9 +30,8 @@ files=("${CONFIG_DIRECTORY:-config}"/crd/bases/*)
 # Copy and package CRDs
 for file in "${files[@]}"; do
   {
-    cat "${file}" | sed -E "/^    controller-gen.kubebuilder.io.version.*/a \\
+    cat "${file}" | sed -e "/^    controller-gen.kubebuilder.io.version.*/a \\
     operator.solr.apache.org\\/version: ${VERSION}\\
-    argocd.argoproj.io\\/sync-options: Replace=true\\
-"
+    argocd.argoproj.io\\/sync-options: Replace=true"
   } > "${file}.tmp" && mv "${file}.tmp" "${file}"
 done
--- a/hack/config/add_crds_annotations.sh
+++ b/hack/config/add_crds_annotations.sh
@@ -21,17 +21,18 @@ set -o pipefail
 # error on unset variables
 set -u
 
-echo "Add headers to CRDs and Role files"
+echo "Add annotations to CRDs"
 
-rm -rf "${CONFIG_DIRECTORY:-config}"/crd/bases/*.tmp "${CONFIG_DIRECTORY:-config}"/rbac/role.yaml.tmp
+rm -rf "${CONFIG_DIRECTORY:-config}"/crd/bases/*.tmp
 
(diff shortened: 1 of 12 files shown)


HoustonPutman:
As I mentioned in the issue, the Solr CRDs are auto-generated by controller-gen, so we need to go through and add the annotations after the generation. (Until controller-gen supports adding CRD annotations)

Will look into how this could work

HoustonPutman:
@janhoy hacked together something that should work for our uses. I also included the version of the operator that the CRDs were made for, which will let users debug and make sure they have the correct CRD installed.

Also I did it in a way that it's done for the helm chart CRD file as well, since users using that are likely to have un-upgraded CRDs installed, so they need the version annotation more than anyone else.

Also included a small fix for the release wizard that I had locally

thelabdude (approved):
very cool solution!

hack/config/add_crds_annotations.sh
```diff
@@ -0,0 +1,37 @@
...
+
+files=("${CONFIG_DIRECTORY:-config}"/crd/bases/*)
+
+# Copy and package CRDs
+for file in "${files[@]}"; do
+  {
+    cat "${file}" | sed -e "/^    controller-gen.kubebuilder.io.version.*/a \\
+    operator.solr.apache.org\\/version: ${VERSION}\\
+    argocd.argoproj.io\\/sync-options: Replace=true"
+  } > "${file}.tmp" && mv "${file}.tmp" "${file}"
```
janhoy, line 36: `sed -i` would do it in-place but may not work on some OSes

janhoy:
Looks good, feel free to take over the PR and merge when done

hack/config/add_crds_annotations.sh
HoustonPutman: Yeah, `sed` differences between OSX and linux drive me insane, especially the `-i`, so it's not used anywhere in the solr-operator scripts. This works for now so I'm happy enough with it, we can always improve it in the future 🙂

HoustonPutman:
The Zookeeper Operator CRD annotations won't be overwritten anymore either, and will be included in the release-artifacts.

PR #413, opened 2022-03-07
Fixes incoherent variable naming for additional domain names

### LINKED ISSUE #412: Incoherent naming for additional domainNames variable (code vs. documentation)
Helm-Chart and CRD documentations mentions a `additionalDomainNames` variable to provide further domainNames.
However, the CRD itself parses additionalDomains only.

mmoscher:
Fixes #412

### DIFF
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -560,7 +560,7 @@ type ExternalAddressability struct {
 	// Provide additional domainNames that the Ingress or ExternalDNS should listen on.
 	// This option is ignored with the LoadBalancer method.
 	// +optional
-	AdditionalDomainNames []string `json:"additionalDomains,omitempty"`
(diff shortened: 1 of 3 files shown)


HoustonPutman:
Unfortunately we can't really break back-compat like that. Either we need to:

- Have 2 fields, and pre-populate the new `additionalDomainNames` field with the old `additionalDomains` value. We can deprecate the `additionalDomains` option and remove it in a future release.
- Fix the documentation to use `additionalDomains` and call it a day.

HoustonPutman:
How about this?

- For `v0.5.1` update the documentation
- For `v0.6.0` add the new field, and auto-populate with the value from the old field. Add upgrade notes, deprecate the old field.
- For `v0.7.0` remove the old field from the CRD

HoustonPutman:
Closing this in favor of #415 and #416

PR #429, opened 2022-04-06
Allow ZkConnectionStrings without ports, log when statefulset is skipped

### LINKED ISSUE #428: Default to port 2181 for zookeeper connection string
The setting `zookeeperConnectionInfo.internalConnectionString` does not have the same semantics as in Solr. It requires an explicit port numer, while in Solr you can say `ZK_HOST=zoo1,zoo2` and it defaults to port 2181.

What happens in solr-operator is that it does not log an understandable error about `internalConnectionString`, instead it refuses to even start the StatefulSet and says something cryptic like this

```
2022-04-06T12:54:05.543Z	ERROR	controller-runtime.manager.controller.solrcloud	Reconciler error	{"reconciler group": "solr.apache.org", "reconciler kind": "SolrCloud", "name": "solr", "namespace": "default", "error": "Operation cannot be fulfilled on solrclouds.solr.apache.org \"solr\": the object has been modified; please apply your changes to the latest version and try again"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:302
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:216
```

The operator should either log the actual error clearly or default to port 2181

HoustonPutman:
Fixes #428


HoustonPutman:
@janhoy , mind giving this a quick look before going on holiday?

controllers/solrcloud_controller.go
```diff
@@ -288,9 +288,11 @@ func (r *SolrCloudReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	// Only create stateful set if zkConnectionString can be found (must contain host and port)
-	if !strings.Contains(newStatus.ZkConnectionString(), ":") {
+	// Only create stateful set if zkConnectionString can be found (must contain a host before the chroot)
+	zkConnectionString := newStatus.ZkConnectionString()
+	if len(zkConnectionString) < 2 || strings.HasPrefix(zkConnectionString, "/") {
 		blockReconciliationOfStatefulSet = true
+		logger.Info("Will not create/update the StatefulSet because the zookeeperConnectionString has no host", "zookeeperConnectionString", zkConnectionString)
```
janhoy, line 295: Should this be a WARN or ERROR? Is is surely not a normal situation?

janhoy (approved):
Have not tested it. Will the string be passed as-is to Solr, i.e. the operator treats it as an opaque value? Then I suppose these simple checks are good to detect empty and malformed string.

Could optionally warn if there is no comma as well, meaning the user has only specified a single Zookeeper instead of three or five?

controllers/solrcloud_controller.go
HoustonPutman: Unfortunately there is no "WARN", but I can change it to an error. It's definitely not a good state to be in.

The string will be passed as-is to Solr. (But there is some validation done, for example the chroot has to be at least a `/`).

I guess we could warn on that, but it'd probably be better to do that warning in Solr vs the Solr Operator.

PR #451, opened 2022-06-30
Ability to customize default initContainer resources

### LINKED ISSUE #448: Allow option to specify init-containers (check-zk, cp-solr-xml etc.) resources when ResourcesQuota is in place
Currently Solr operator doesn't allow setting init-containers resource limits and resulting into Solr pods to fail when ResourceQuota is in place.  

Error
===
create Pod wm-solrcloud-0 in StatefulSet wm-solrcloud failed error: pods "wm-solrcloud-0" is forbidden: failed quota: aes-wm-dit: must specify limits.cpu,limits.memory,requests.cpu,requests.memory

HoustonPutman:
Resolves #448 

The new SolrCloud option is `Spec.customSolrKubeOptions.podOptions.defaultInitContainerResources`.

The option can also be enabled in the Solr Helm chart through: `podOptions.defaultInitContainerResources`.


controllers/util/solr_util.go
```diff
@@ -607,12 +615,20 @@ func generateSolrSetupInitContainers(solrCloud *solr.SolrCloud, solrCloudStatus
 		}
 	}
 
+	volumePrepResources := corev1.ResourceList{
+		corev1.ResourceCPU:    *DefaultSolrVolumePrepInitContainerCPU,
```
huijieliu, line 619: 👏This will help to make sure if we configure our solr resources with limit=request as well,  the pod will get assigned with QoS class Guaranteed

helm/solr/templates/_custom_option_helpers.tpl
```diff
@@ -98,6 +98,10 @@ initContainers:
 topologySpreadConstraints:
   {{- toYaml .Values.podOptions.topologySpreadConstraints | nindent 2 }}
 {{ end }}
+{{- if .Values.podOptions.defaultInitContainerResources -}}
+defaultInitContainerResources:
```
huijieliu, line 102: discussed offline, using default resource to configure both init container resource should be good for now if we need to custom init container resources as both are light weight init containers.

huijieliu (commented):
Thanks for making this change!

api/v1beta1/common_types.go
```diff
@@ -150,6 +149,10 @@ type PodOptions struct {
 	// +listMapKey=whenUnsatisfiable
 	// +optional
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// ProvidedInitContainerResources is the resource requirements for the default init container(s) created by the Solr Operator if any are created.
```
gerlowskija, line 153: [0] Looks like this comment is out of sync with the new field-name you ultimately went with here?

gerlowskija (approved):
LGTM!

api/v1beta1/common_types.go
HoustonPutman: Good catch!

PR #452, opened 2022-07-05
Make new TLSTermination option, with or w/o secret

### LINKED ISSUE #437: Generate tls: entry for ingress entries even without a tls secret
The nginx ingress controller supports [using a default certificate for connections that don't have an explicit secret set](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#default-ssl-certificate) in their `tls` section. This default certificate is only used if a `secretName` is not provided. 

As far as I've been able to tell [from how the Ingress entry is created](https://github.com/apache/solr-operator/blob/395e83861cfe6bdb31dd3f1dcc94847af41ba01e/controllers/util/solr_util.go#L862) there is no way of getting an ingressTLS object added without providing some sort of secret.

Would it be possible to get an option to make the operator create an ingress entry with TLS, but without a secret name to make nginx use the default certificate? (which usually will be a wildcard certificate matching the at a certain level of the domainName given).

HoustonPutman:
Resolves #437

- Use old option to default new `Spec.solrAddressability.external.ingressTLSTermination.tlsSecret`.
- Deprecate old `Spec.solrAddressability.external.ingressTLSTerminationSecret`.

Things to do:

- [x] Documentation
- [x] Upgrade Notes
- [x] Tests for the deprecated option migration
- [x] Tests for the new functionality


api/v1beta1/solrcloud_types.go
```diff
@@ -576,8 +588,21 @@ const (
...
 func (opts *ExternalAddressability) withDefaults(usesTLS bool) (changed bool) {
+	// TODO: Remove in v0.7.0
+	// If the deprecated IngressTLSTerminationSecret exists, use it to default the new location of the value.
+	// If that location already exists, then merely remove the deprecated option.
+	if opts.IngressTLSTerminationSecret != "" {
+		if !opts.HasIngressTLSTermination() {
+			opts.IngressTLSTermination = &SolrIngressTLSTermination{
+				TLSSecret: opts.IngressTLSTerminationSecret,
+			}
+		}
```
gerlowskija, line 599: [Q] If we receive a solrcloud that has both the deprecated IngressTLSTerminationSecret and an IngressTLSTermination object, should we log a warning or note the conflict in some other way?

I guess we wouldn't expect users to necessarily be checking the operator logs for that sort of thing.  So maybe there's no good way to surface this.  Just thinking aloud I guess...

gerlowskija (commented):
LGTM, pending the items in the TODO list above.

api/v1beta1/solrcloud_types.go
HoustonPutman: That's a good idea.

HoustonPutman: Going to tackle this tomorrow. Everything else should be done though

HoustonPutman: Done!

PR #455, opened 2022-07-07
Recompute NextScheduled backup on each reconcile

### LINKED ISSUE #376: Recompute solrbackup's `.Status.NextScheduledTime` on changes to `.Spec.Recurrence.Schedule`
Currently, changing the `schedule` property leaves the "Next Scheduled" time untouched, which causes a delay in when the `schedule` property is actually able to "take effect".

e.g. Changing `Schedule` from "@every 1w" to "@every 1d" will increase the frequency, but not until "NextScheduledTime" comes around (which may be up to a week away!).

Exactly how the "Next Scheduled Time" should be recomputed is a good question that I can see arguments on either side of.  e.g. Continuing the example above: should the next backup be one day from the previous backup?  One day from the current time?  Something else?

But regardless of opinions there, any of those options seem preferable to not recomputing NextScheduledTime at all.

gerlowskija:
Resolves #376

Prior to this commit, the 'NextScheduled' field of a solrbackup's status
was never updated, even if the user the backup schedule in the
solrbackup spec.

This commit changes this: NextScheduled is now recomputed during each
reconciliation, as long as there is not an active backup ongoing.


HoustonPutman:
We were discussing offline about how the schedule gets restarted if the user disables the recurrence and then re-enables it. Does this fix that?

HoustonPutman (approved):
Otherwise this looks simple and straightforward, LGTM!

gerlowskija:
> how the schedule gets restarted if the user disables the recurrence and then re-enables it

This fixes that "problem", yep.  Prior to this PR, a backup would immediately be run upon re-enabling the `solrbackup`.  With this PR, we now wait if the time indicated by `NextScheduled` hasn't arrived yet.  (Of course, if NextScheduled has already come and gone when a solrbackup is re-enabled, we _will_ trigger a backup upon re-enabling.)

PR #461, opened 2022-07-28
Option to merge the JVM truststore with user-supplied truststore

### LINKED ISSUE #390: auto certificate renewal with restartOnTLSSecretUpdate and cert-manager fails
Asked in https://kubernetes.slack.com/archives/CQSNS615F/p1639651250121300, recommended to bring here.

I’m seeing an issue where my Solr cloud (3 replicas) doesn’t recover when cert-manager renews the certificate. It seems to be tied to my updateStrategy.
What I’m seeing:

- Certificate is renewed by cert-manager at 60days (expiry is 90days - default for LetsEncrypt), `restartOnTLSSecretUpdate` setting starts to restart the cluster

```yaml
solrTLS:  
    restartOnTLSSecretUpdate: true
    pkcs12Secret:
      name: solr-tls
      key: keystore.p12
    keyStorePasswordSecret:
      name: pkcs12-keystore-password
      key: password-key
```

- Below settings ensure only 1 pod restarts at a time so we still have 2 active pods to serve requests

```yaml
  updateStrategy:
    managed:
      maxPodsUnavailable: 1
      maxShardReplicasUnavailable: 1
```

After the 1 Pod restarts, its collections are shown as having a “Down” status, in the logs I see

```yaml
2021-12-16 10:15:14.746 ERROR (recoveryExecutor-11-thread-3-processing-n:pod-1.default:443_solr x:transaction_10_shard1_replica_n5 c:transaction_10 s:shard1 r:core_node6) [c:transaction_10 s:shard1 r:core_node6 x:transaction_10_shard1_replica_n5] o.a.s.c.RecoveryStrategy Failed to connect leader https://pod-2.default:443/solr on recovery, try again
```

- The cluster is stuck at this state, as the updateStrategy settings won’t allow the other 2 pods to restart since the 1 pod’s collections are “Down”
The error seems to indicate that the pod that has restarted can’t communicate with the pods that haven’t restarted yet, presumably because the certificates.

- A (bad) “workaround” I found is that if it set the below, then all 3 pods will restart and come up heathy. But this obviously has the downside of a short downtime.

```yaml
updateStrategy:
    managed:
      maxPodsUnavailable: 3
      maxShardReplicasUnavailable: 3
```

I don’t think the active certs would have been expired, as Lets Encrypt certs have 90day duration and they renewed at 60days (which is the default for cert-manager of 2/3 of the duration). https://cert-manager.io/docs/usage/certificate/#renewal & https://cert-manager.io/docs/faq/#if-renewbefore-or-duration-is-not-defined-what-will-be-the-default-value

I wonder if the act of cert-manager renewing the certificates invalidates the active one though? I can’t find this specifically in the docs. This would be a problem then. It also would explain why I was still seeing this problem when triggering a manual renewal using the cert-manager cmctl https://cert-manager.io/docs/usage/cmctl/#renew cli. If this is the case, we would need `restartOnTLSSecretUpdate` to be able to ignore the `updateStrategy. managed. max*Unavailable` settings.

Sort of related issue: https://github.com/jetstack/cert-manager/issues/1168, but Solr has `restartOnTLSSecretUpdate` for our pods.

thelabdude:
Fixes #390 ~ by allowing the JVM cacerts to get merged in with the user-supplied truststore (Let's Encrypt's CA is in the cacerts for modern JVM)

Users can now configure the TLS options to merge the JVM's truststore with the truststore for their certs using:
```
spec:
  ...
  solrTLS:
    ...
    trustStoreSecret:
      name: dev-selfsigned-cert-tls
      key: truststore.p12
    mergeJavaTrustStore: "$JAVA_HOME/lib/security/cacerts"
```
The path given in `mergeJavaTrustStore` option must exist on the Solr docker image! Thus, if user's customize their Solr image, this path may be different.

Behind the scenes, this creates an additional `initContainer` that merges the two truststores together and then points the env var to the "merged" truststore:

For server TLS:
```
- name: SOLR_SSL_TRUST_STORE
   value: /var/solr/tls-merged/truststore.p12
``

By pointing `SOLR_SSL_TRUST_STORE` env var at the merged truststore, we're ensured that all the other initContainers and liveness probes continue to work (as they just use the env var to resolve this path).

Added a few simple unit tests and tested manually in my cluster. 

For Prom exporter, the config would be:
```
spec:
  solrReference:
    ...
    solrTLS:
      ...
      mergeJavaTrustStore: "$JAVA_HOME/lib/security/cacerts"
```

Which results in the exporter container getting configed with env var:
```
- name: SOLR_SSL_CLIENT_TRUST_STORE
   value: /var/solr/tls-merged/truststore.p12
```

### DIFF
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -1523,6 +1523,14 @@ type SolrTLSOptions struct {
 	// This option is typically used with `spec.updateStrategy.restartSchedule` to restart Solr pods before the mounted TLS cert expires.
 	// +optional
 	MountedTLSDir *MountedTLSDirectory `json:"mountedTLSDir,omitempty"`
+
+	// Path on the Solr image to your JVM's truststore to merge with an external truststore.
+	// If supplied, Solr will be configured to use the merged truststore.
+	// The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts
+	MergeJavaTruststore string `json:"mergeJavaTrustStore,omitempty"`
+
+	// Password for the Java truststore to merge; defaults to "changeit"
+	MergeJavaTruststorePass string `json:"mergeJavaTrustStorePass,omitempty"`
 }
 
 // +kubebuilder:validation:Enum=Basic
--- a/config/crd/bases/solr.apache.org_solrclouds.yaml
+++ b/config/crd/bases/solr.apache.org_solrclouds.yaml
@@ -4926,6 +4926,12 @@ spec:
                     required:
                     - key
                     type: object
+                  mergeJavaTrustStore:
+                    description: 'Path on the Solr image to your JVM''s truststore to merge with an external truststore. If supplied, Solr will be configured to use the merged truststore. The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts'
+                    type: string
+                  mergeJavaTrustStorePass:
+                    description: Password for the Java truststore to merge; defaults to "changeit"
+                    type: string
                   mountedTLSDir:
                     description: Used to specify a path where the keystore, truststore, and password files for the TLS certificate are mounted by an external agent or CSI driver. This option is typically used with `spec.updateStrategy.restartSchedule` to restart Solr pods before the mounted TLS cert expires.
                     properties:
@@ -5087,6 +5093,12 @@ spec:
                     required:
                     - key
                     type: object
+                  mergeJavaTrustStore:
+                    description: 'Path on the Solr image to your JVM''s truststore to merge with an external truststore. If supplied, Solr will be configured to use the merged truststore. The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts'
+                    type: string
+                  mergeJavaTrustStorePass:
+                    description: Password for the Java truststore to merge; defaults to "changeit"
+                    type: string
                   mountedTLSDir:
                     description: Used to specify a path where the keystore, truststore, and password files for the TLS certificate are mounted by an external agent or CSI driver. This option is typically used with `spec.updateStrategy.restartSchedule` to restart Solr pods before the mounted TLS cert expires.
                     properties:
--- a/config/crd/bases/solr.apache.org_solrprometheusexporters.yaml
+++ b/config/crd/bases/solr.apache.org_solrprometheusexporters.yaml
@@ -3688,6 +3688,12 @@ spec:
                         required:
                         - key
                         type: object
+                      mergeJavaTrustStore:
+                        description: 'Path on the Solr image to your JVM''s truststore to merge with an external truststore. If supplied, Solr will be configured to use the merged truststore. The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts'
+                        type: string
+                      mergeJavaTrustStorePass:
+                        description: Password for the Java truststore to merge; defaults to "changeit"
+                        type: string
                       mountedTLSDir:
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Review by HoustonPutman — CHANGES_REQUESTED** [review]
Would you be able to add something to the docs about this as well?

Also a changelog entry would be much appreciated 🙂

File: api/v1beta1/solrcloud_types.go [hunk 1]
```diff
@@ -1523,6 +1523,14 @@ type SolrTLSOptions struct {
...
 	// +optional
 	MountedTLSDir *MountedTLSDirectory `json:"mountedTLSDir,omitempty"`
+
+	// Path on the Solr image to your JVM's truststore to merge with an external truststore.
+	// If supplied, Solr will be configured to use the merged truststore.
+	// The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts
+	MergeJavaTruststore string `json:"mergeJavaTrustStore,omitempty"`
+
+	// Password for the Java truststore to merge; defaults to "changeit"
+	MergeJavaTruststorePass string `json:"mergeJavaTrustStorePass,omitempty"`
```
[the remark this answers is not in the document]
thelabdude: seemed like overkill to me since it's the truststore pass for the JVM which is most likely "changeit" and isn't used by Solr ... that said, we can pull from a secret too ... doubt many would ever use this option.

api/v1beta1/solrcloud_types.go
```diff
@@ -1523,6 +1523,14 @@ type SolrTLSOptions struct {
 	// This option is typically used with `spec.updateStrategy.restartSchedule` to restart Solr pods before the mounted TLS cert expires.
 	// +optional
 	MountedTLSDir *MountedTLSDirectory `json:"mountedTLSDir,omitempty"`
+
+	// Path on the Solr image to your JVM's truststore to merge with an external truststore.
+	// If supplied, Solr will be configured to use the merged truststore.
+	// The truststore for the JVM in the default Solr image is: $JAVA_HOME/lib/security/cacerts
+	MergeJavaTruststore string `json:"mergeJavaTrustStore,omitempty"`
```
HoustonPutman, line 1530: Does this work with `mountedTLSDir`?

api/v1beta1/solrcloud_types.go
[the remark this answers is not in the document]
HoustonPutman: ahhh ok, didn't understand it was unusual to change it. Sounds good

api/v1beta1/solrcloud_types.go
thelabdude: Typically, `mountedTLSDir` will have a csi driver volume and corresponding mount on the mainContainer, which would get used by the merge `initContainer`, so the init container would get the truststore file. However, that might cause double creation of the cert for each pod, once for the `initContainer` and once for the main container, so this would likely put a lot of pressure on the Cert issuer. So probably safer to say this feature is not supported with `mountedTLSDir` option for now.

thelabdude: No, nvm all that `volumesAndMounts` doesn't include vols & mounts for the `mountedTLSDir` option. So this option is not going to work with `mountedTLSDir`.

So maybe we just punt on this feature for 0.6 and solve it using an init-db script instead? There's already some code in place for mounting a script into the init-db.

thelabdude:
I think we should hold off on this solution for 0.6 as now I'm thinking the better approach would be to use the init-db script approach built into the Solr image. The problem with the current solution in this PR is it doesn't work with the `mountedTLSDir` option. Originally I was thinking the volumes and mounts would be available to the merge initContainer but that's not the case.  I think the init-db approach will actually be fairly easy to implement except for the Solr exporter. Unfortunately, I won't have time to work on this for a bit and I don't want to hold up 0.6

HoustonPutman:
Sounds good Tim. Thanks for putting in the effort on this though!

PR #462, opened 2022-08-05
Use GOARCH=amd64 when testing on 'Apple Silicon'

gerlowskija:
There is no arm64 artifact for our current version of kubebuilder-tools,
which causes test setup to fail on M1 machines.  However, the amd64
artifact offered by kubebuilder seems to work well enough for our
purposes, so we can use that as a workaround until we're able to upgrade
kubebuilder-tools down the road.


HoustonPutman (approved):
Looks good!

PR #463, opened 2022-08-06
Refactor helm logic in test_cluster.sh

gerlowskija:
This fixes a bug in the smoketester wherein we attempted to 'upgrade' a
helm package that had previously been removed during some mid-execution
cleanup.


HoustonPutman (approved):
Looks good to me!

PR #464, opened 2022-08-12
Update local_tutorial.md

krickert:
* Changed docker-edge to docker.  docker for mac no longer has the docker-edge version.  Now, just need to have "docker".  
* changed "brew cask" command to "brew --cask", which is the newer syntax that works with brew.  "brew cask" is no longer supported.


janhoy (approved):

janhoy:
Thanks for the contribution!

HoustonPutman:
Thanks for the improvement!

In the future, we also need to update the `main` branch, since this gets overridden with whatever the release documentation is, when a release happens. I will go ahead and port this to `main` and `release-0.6` as well!

janhoy:
Ah, need to dig into the docs workflow. Thanks Houston.

PR #476, opened 2022-09-18
Repeating Backup Solr Version is confusing

orangewolf:
Previous version implied solr had to be less than 8.9.0, but incremental backup is introduced in 8.9.0. Instead of “or older” vs “or newer” >= 8.9.0 is used, as this is more clear for non native English speakers and matches other version info in the document.


HoustonPutman:
Thanks, documentation PRs are always much appreciated! (Next time you can make the PR against `main`, and we will handle backporting to `gh-pages` 🙂)

PR #480, opened 2022-10-05
Add Annotations to created ServiceAccount

endzyme:
In order to support AWS IAM Role assumption via the [pod-identity-webhook](https://github.com/aws/amazon-eks-pod-identity-webhook), there need to be annotations added to the ServiceAccount.

This change enables users to add annotations to the ServiceAccount created for the SolrCloud instance.

### DIFF
--- a/helm/solr-operator/Chart.yaml
+++ b/helm/solr-operator/Chart.yaml
@@ -61,9 +61,9 @@ annotations:
         - name: Github PR
           url: https://github.com/apache/solr-operator/pull/481
     - kind: added
-      description: ServiceAccount annotations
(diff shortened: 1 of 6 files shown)


HoustonPutman (commented):
This looks good to me.

Could you add a changelog entry in the Chart.yaml, and add it to the list of available options in the chart's README?

This is also something that we could add to the solr-operator chart.

endzyme:
@HoustonPutman I've updated the changelog and added the same feature to the solr-operator chart.

HoustonPutman:
Made a fix for the serviceAccount when annotations are not provided, but otherwise should be good to go! This will make it into `v0.7.0`

PR #485, opened 2022-11-01
Support for Projected Volume

### LINKED ISSUE #484: Support Projected Volume in Solr Pod spec
We can mount volumes on the Solr pod today, but only plain ones.

Need support for [Projected Volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/), which would look like this in Helm speak:

```yaml
solr:
  podOptions:
    volumes:
      # Map service account JWT token onto a volume
      - name: sa-token-vol
        defaultContainerMount:
          mountPath: /var/run/secrets/tokens
          name: sa-token-vol
        projected:
          sources:
            - serviceAccountToken:
                path: sa-token
                expirationSeconds: 86400
                audience: myAudience
```

This example will cause k8s to generate a Service Account token in `/var/run/secrets/tokens/sa-token` and rotate it automatically. Then Solr (or some plugin within Solr) can fetch the token from that file and use it to authenticate with other services within the cluster. We need this for a custom `AuditPlugin` which will post Audit Events to another service in the cluster, which requires ServiceAccount tokens.

Today [AdditionalVolume](https://github.com/apache/solr-operator/blob/main/api/v1beta1/common_types.go#L204-L217) only accepts plain Volume mounts.

janhoy:
This is my first attempt at a contribution to the operator, implementing #484. But I cannot get the debugger working in IDEA so I'm not sure where it goes wrong.

Fixes #484

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -207,13 +207,16 @@ type AdditionalVolume struct {
 	Name string `json:"name"`
 
 	// Source is the source of the Volume to be loaded into the solrCloud Pod
-	Source corev1.VolumeSource `json:"source"`
+	Source corev1.VolumeSource `json:"source,omitempty"`
 
 	// DefaultContainerMount defines how to mount this volume into the default container.
 	// If this volume is to be used only with sidecar or non-default init containers,
 	// then this option is not necessary.
 	// +optional
 	DefaultContainerMount *corev1.VolumeMount `json:"defaultContainerMount,omitempty"`
+
+	// Projected Volume source
+	ProjectedSource *corev1.ProjectedVolumeSource `json:"projected,omitempty"`
 }
 
 // ContainerImage defines the fields needed for a Docker repository image. The
--- a/controllers/controller_utils_test.go
+++ b/controllers/controller_utils_test.go
@@ -856,6 +856,32 @@ var (
 			},
 		},
 	}
+	// Volume Projection test
+	expiration = int64(300)
+	mySource   = []corev1.VolumeProjection{
+		{
+			ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+				Audience:          "My.Audience",
+				Path:              "token",
+				ExpirationSeconds: &expiration,
+			},
+		},
+	}
+	extraProjectedVolumes = []solrv1beta1.AdditionalVolume{
+		{
+			Name: "projVol",
+			ProjectedSource: &corev1.ProjectedVolumeSource{
+				Sources:     mySource,
+				DefaultMode: nil, // TODO: Use other value?
+			},
+			DefaultContainerMount: &corev1.VolumeMount{
+				Name:      "ignore",
+				ReadOnly:  false,
+				MountPath: "/test/mount/path",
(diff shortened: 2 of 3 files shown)


janhoy:
The test output is

```
• Failure [1.304 seconds]
SolrCloud controller - General
/Users/janhoy/git/solr-operator/controllers/solrcloud_controller_test.go:35
  Solr Cloud with Projected Volume
  /Users/janhoy/git/solr-operator/controllers/solrcloud_controller_test.go:707
    has the correct resources [It]
    /Users/janhoy/git/solr-operator/controllers/solrcloud_controller_test.go:725

    Additional Volume from podOptions not loaded into pod properly.
    Expected
        <*v1.ProjectedVolumeSource | 0x0>: nil
    to equal
        <*v1.ProjectedVolumeSource | 0x106e1c780>: {
            Sources: [
                {
                    Secret: nil,
                    DownwardAPI: nil,
                    ConfigMap: nil,
                    ServiceAccountToken: {
                        Audience: "My.Audience",
                        ExpirationSeconds: 300,
                        Path: "token",
                    },
                },
            ],
            DefaultMode: nil,
        }

    /Users/janhoy/git/solr-operator/controllers/solrcloud_controller_test.go:736
```

janhoy:
Digging some more, I see that in the spec, `corev1.ProjectedVolumeSource` is alredy a child of `corev1.VolumeSource`, so perhaps this already is supported by nesting `projected` inside `source` instead of the shortcut form documented in https://kubernetes.io/docs/concepts/storage/projected-volumes/#example-configuration-secret-downwardapi-configmap?

```yaml
solr:
  podOptions:
    volumes:
      # Map service account JWT token onto a volume
      - name: sa-token-vol
        defaultContainerMount:
          mountPath: /var/run/secrets/tokens
          name: sa-token-vol
        source:
          projected:
            sources:
              - serviceAccountToken:
                  path: sa-token
                  expirationSeconds: 86400
                  audience: myAudience
```

HoustonPutman:
> Digging some more, I see that in the spec, `corev1.ProjectedVolumeSource` is alredy a child of `corev1.VolumeSource`

Yeah it looks like that is the case! It sounds like you had a use case for it, so do you want to give it a try and see how it works?

Looks like its there in the documentation, so this isn't a thing in the v1.20 Kube APIs (the version of the code dependency that we are using) that was removed in a later version.

HoustonPutman:
But your PR is quite close, you've done a good job!

You would have had two steps left:
- Actually set the projected volume on the pods in the controller (you have just created a CRD entry and test, the test fails because you aren't actually using the information from the CRD to set the Pod).
  - https://github.com/apache/solr-operator/blob/main/controllers/util/solr_util.go#L237
  - https://github.com/apache/solr-operator/blob/main/controllers/util/prometheus_exporter_util.go#L158
- Run `make prepare` to do the code/config generation and auto format everything.

janhoy:
I tested the yaml above and it actually works - I now have the `/var/run/secrets/tokens/sa-token` file in my pod! Closing this. Perhaps I'll add another PR to document how to mount a projected volume.

PR #494, opened 2022-11-07
Upgrade go to 1.19, enforce in Makefile

### LINKED ISSUE #492: Upgrade GoLang version to v1.19
Note: We should have the Makefile fail immediately on an unsupported go version.

HoustonPutman:
Fixed #492 

Required by #491


HoustonPutman:
I have to upgrade controller-gen because of incompatibilities with go 1.19, but it will be further updated in #491

Makefile
```diff
@@ -18,6 +18,13 @@ PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
+GO_VERSION = $(shell go version | sed -r 's/^.*([0-9]+\.[0-9]+\.[0-9]+).*$$/\1/g')
+REQUIRED_GO_VERSION = $(shell cat go.mod | grep -E 'go [1-9]\.[0-9]+' | sed -r 's/^go ([0-9]+\.[0-9]+)$$/\1/g')
+
+ifeq (,$(findstring $(REQUIRED_GO_VERSION),$(GO_VERSION)))
```
gerlowskija, line 24: [+1] I remember we talked about this being one of the reasons people ahve trouble running the operator smoke-tester and build, awesome to add a check in here for it.

config/crd/bases/solr.apache.org_solrbackups.yaml
```diff
@@ -63,10 +63,14 @@ spec:
         description: SolrBackup is the Schema for the solrbackups API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
```
gerlowskija, line 66: [-0] Ugh, these reformatting changes are unfortunate.  No way around them I take it?

gerlowskija (approved):
LGTM.

The yaml reformatting, which looks like it comes from a difference in the acceptable max-line-length, is unfortunate.  But I imagine there was no way around it.

config/crd/bases/solr.apache.org_solrbackups.yaml
HoustonPutman: No workaround. It'll probably get fixed when we upgrade the controller-gen version, but we will see.

PR #496, opened 2022-11-08
Upgrade Ginkgo and EnvTest

HoustonPutman:
With upgrading EnvTest, we have a much easier setup for tests!

Ginkgo we are upgrading to v2, which has small back-incompatabilities.


HoustonPutman:
@janhoy and @gerlowskija I know y'all have both had issues with your M1 macbooks and the build system before. So I definitely want to have y'all successfully do a run through before merging this 🙂

HoustonPutman:
Note, when you upgrade your golang version (1.19 is now required) I highly suggest running `make clean mod-tidy` to clear all of your cached dependencies

gerlowskija:
This works for me!  I had a small issue with bad syntax on a `chmod` command used by `make clean`.  (I've pushed that fix up to this PR after a short conversation offline with Houston)

gerlowskija (approved):

PR #499, opened 2022-11-23
Fix typo in readme

janhoy:
Also specify that changing to another version of the zookeeper image will not work (since zookeeper-operator [hard-codes the start script to `/usr/local/bin/zookeeperStart.sh`](https://github.com/pravega/zookeeper-operator/blob/master/pkg/zk/generators.go#L156) and that script is pravega-specific and does not exist in e.g. the official zookeeper image.


HoustonPutman (approved):
Good catch!

HoustonPutman:
Want to merge this @janhoy ?

PR #507, opened 2022-12-16
Start e2e/integration testing functionality.

### LINKED ISSUE #134: Introduce comprehensive integration tests
Our unit tests are quite good for a wide variety of use cases, however operator interacts with external systems quite often now.

This includes (but is not limited to):
- Creating ZK Nodes for SolrCloud chRoots.
- Calling Solr APIs when dealing with SolrCollections, SolrCollectionAliases and SolrBackups.
- The new way of upgrading solr pods in #133 

These are all use cases that cannot be adequately tested in unit tests. Right now it's relatively easy for us to manually test changes to these parts of the operator, however this reliance on manual tests will not scale to the future when the operator has much more responsibility.

This issue is just to start the discussion, I haven't found any good frameworks to get this started with yet. Hopefully we can put our heads together and come up with something great! 😄

HoustonPutman:
Resolves #134

This is the first of hopefully many integration testing tickets, mostly to build the start of a framework so that other work can be done iteratively.

Goals/Progress:

- [x] Integration testing with the Solr Operator Helm Chart
- [x] At least 1 integration test for each Solr resource
- [x] Documentation
- [x] Good cleanup while retaining ability to debug when things didn't work
- [x] The go code should not be calling external programs. See usage of Helm golang APIs.

Future Goals, not required for this ticket:

- [ ] TLS testing
- [ ] Being able to use this component for the release smoke tests
- [ ] Well supported ability to use a non-KiND cluster
- [ ] Be able to run without access to the internet


HoustonPutman:
Ok smoketester usage is now completely covered by the e2e tests. This should be pretty close to merging in early january once we have some other eyes on it!

Btw https://github.com/apache/solr-operator/pull/511 is required for the tests to pass, so we should decide on that before merging this in.

HoustonPutman:
Ok, big win. Much like the solr tests, the e2e tests are now completely reproducable with all possible inputs (random seed, operator/solr image, kubernetes version, etc), and accept easy filtering.

On test failure each individual test will also output a command to reproduce the test failure.

Screen Shot 2023-01-05 at 3 41 05 PM

This should be ported to the unit tests in a separate PR, because its an awesome feature and very much missing from the operator tests.

We are reaching the finish line here. I am going to document the tests and clean up `e2e-tests` and `e2e-tests-run`, so that there is only 1 task. Then I think this will be ready to merge, hopefully next week.

HoustonPutman:
Ok I don't think Github Action runners have the resources to run the e2e tests, so we will have to disable those for now. We can have them run in Jenkins instead for now.

HoustonPutman:
I think this is ready to go and plan to commit it early next week!

PR #509, opened 2022-12-21
Fix non-recurring backups

HoustonPutman:
https://github.com/apache/solr-operator/pull/455 introduced a bug for non-recurring backups that was unearthed while working on https://github.com/apache/solr-operator/pull/507

Basically we need to only update the `NextScheduledTimestamp` if `recurrence` is enabled.

I also restructured the logic to hopefully make it more clear when backup logic should be run.

### DIFF
--- a/controllers/solrbackup_controller.go
+++ b/controllers/solrbackup_controller.go
@@ -92,6 +92,7 @@ func (r *SolrBackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	requeueOrNot := reconcile.Result{}
 
+	// Backup work needs to be done by default if the current backup is not finished
 	doBackupWork := !backup.Status.IndividualSolrBackupStatus.Finished
 
 	// Check if we should start the next backup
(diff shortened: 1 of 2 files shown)


controllers/solrbackup_controller.go
```diff
@@ -146,7 +145,7 @@ func (r *SolrBackupReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	// Schedule the next backupTime, if it doesn't have a next scheduled time, it has recurrence and the current backup is finished
-	if backup.Status.IndividualSolrBackupStatus.Finished {
+	if backup.Status.IndividualSolrBackupStatus.Finished && backup.Spec.Recurrence.IsEnabled() {
```
gerlowskija, line 148: [+1] Weird that the comment on L147 mentions recurrence-being-enabled as being a prereq, but we never checked for it before now.  Must've just been an oversight on my part?

Anyway, good catch!

gerlowskija (approved):
LGTM - maybe next time we could put some of the more cosmetic improvements in their own commit, but it seems pretty harmless here.

Thanks for catching my bug!

HoustonPutman:
> maybe next time we could put some of the more cosmetic improvements in their own commit, but it seems pretty harmless here.

Good call, just included since I was going mad trying to fix it so many ways before realizing that I wasn't testing the changes at all....

PR #514, opened 2022-12-30
Change Zookeeper Cluster CRD Spec Labels

endzyme:
Zookeeper cluster CRD spec.pod.labels presently does not function as intended. StatefulSet pods receive all their labels from spec.labels. This is outlined in https://github.com/pravega/zookeeper-operator/issues/511.

This change makes it so that any labels which are provided in the pod labels are merged into the spec.labels of the Zookeeper cluster CRD. This also prioritizes any labels provided by the Solr Operator, thus not allowing you to override any important labels like 'app' or 'release' when using the spec.pod.labels in the SolrCloud CRD.


HoustonPutman (approved):
Yeah this looks good, their CRD is kind of a mess with some pod options living under `Spec.pod`, and some just living under `Spec`...

Thanks for adding the tests as well!

I'll get this merged if you can add a quick changelog entry (you can use the other changes as a template for yours):
https://github.com/apache/solr-operator/blob/main/helm/solr-operator/Chart.yaml#L55

endzyme:
Thanks again!

PR #516, opened 2023-01-12
Fix SolrBackup not taking backups when the collections field is omitted 

### LINKED ISSUE #515: SolrBackup doesn't work when the collections field is omitted
The SolrBackup CRD has the option to specify collections for backups. However, it's not a required field. One could imagine that without specifying this field, every collection should be backed up.

On version 0.6.0 when the collection field is not configured, nothing happens within the controller, only this is being logged.
```
2023-01-12T12:10:37.412Z	INFO	controller-runtime.manager.controller.solrbackup	Updating status for solr-backup	{"reconciler group": "solr.apache.org", "reconciler kind": "SolrBackup", "name": "test-whole-backup", "namespace": "one-preprod", "newStatus": {"solrVersion":"8.11.2","startTimestamp":"2023-01-12T12:10:37Z"}, "oldStatus": {"startTimestamp":null}}
```
Solr doesn't receive any API calls to backup every collection.

a-nych:
Resolves #515.

When there are no collections specified in the CRD, attempt to list them all from SolrCloud and back them up.

### DIFF
--- a/controllers/solrbackup_controller.go
+++ b/controllers/solrbackup_controller.go
@@ -223,8 +223,18 @@ func (r *SolrBackupReconciler) reconcileSolrCloudBackup(ctx context.Context, bac
 		currentBackupStatus.StartTime = metav1.Now()
 	}
 
+	collectionsToBackup := backup.Spec.Collections
+
+	// If there are no collections specified, we need to list through collections available in Solr
+	if len(collectionsToBackup) == 0 {
+		collectionsToBackup, err = util.ListAllSolrCollections(ctx, solrCloud, logger)
+		if err != nil {
+			logger.Error(err, "Error listing collections", "solrCloud", solrCloud.Name)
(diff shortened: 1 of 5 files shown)


HoustonPutman (approved):
This looks great to me! Thanks for the great contribution 🙂 

One request, could you add a changelog entry [here](https://github.com/apache/solr-operator/blob/main/helm/solr-operator/Chart.yaml#L53)? Follow the syntax use by the other entries.

Btw it will be great to be able to test this when https://github.com/apache/solr-operator/pull/507 is merged!

a-nych:
@HoustonPutman Thanks for the feedback! I've added the changelog. 

Do we want to wait for #507 before merging this?

HoustonPutman:
This looks good to me! We can merge this before the e2e tests, and add a test for it after that is merged!

PR #520, opened 2023-01-23
Use correct user for chmod command in cp-solr-xml init container

### LINKED ISSUE #519: cp-solr-xml fails when a local backup repository is configured with SolrCloud running as a custom user
Steps to reproduce:

1. Run SolrCloud nodes with `podOptions.podSecurityContext.runAsUser` and `podOptions.podSecurityContext.fsGroup` set to anything other than `8983` (default SolrCloud user)
2. Configure a local `backupRepository` using `persistentVolumeClaim`

`cp-solr-xml` will then run `chmod -r 8983:8983 <backup_storage_path>` despite us running Solr with different users. At some point when files are created in the backup repository by SolrCloud, the `chmod` command will fail due to missing permissions and SolrCloud won't boot.

a-nych:
Resolves #519.


controllers/util/solr_util.go
```diff
@@ -594,6 +594,21 @@ func generateSolrSetupInitContainers(solrCloud *solr.SolrCloud, solrCloudStatus
 	}
 	setupCommands := []string{"cp /tmp/solr.xml /tmp-config/solr.xml"}
 
+	// Figure out the solrUser and solrGroup to use
+	solrUser := DefaultSolrUser
+	solrGroup := DefaultSolrGroup
+	solrPodSecurityContext := solrCloud.Spec.CustomSolrKubeOptions.PodOptions.PodSecurityContext
+
+	if solrPodSecurityContext.RunAsUser != nil {
```
HoustonPutman, line 602: `solrPodSecurityContext` might be nil, which is why the tests are failing.

HoustonPutman (commented):
So I have a general question about this. We only specify the `fsGroup` [by default](https://github.com/apache/solr-operator/blob/main/controllers/util/solr_util.go#L479). Do we not want to use the `fsGroup` provided in the custom `PodSecurityContext`?

controllers/util/solr_util.go
```diff
@@ -594,6 +594,21 @@ func generateSolrSetupInitContainers(solrCloud *solr.SolrCloud, solrCloudStatus
...
+		solrUser = int(*solrPodSecurityContext.RunAsUser)
+
+		if solrPodSecurityContext.RunAsGroup != nil {
```
HoustonPutman, line 605: The user might specify a `runAsGroup` without a `runAsUser`, so we probably want to keep these two checks separate right? Or am I thinking about this incorrectly?

HoustonPutman:
@a-nych I'm pushing for a v0.7.0 release soon and it would be great to get this in there. Do you have any more info for my `runAsUser` vs `runAsGroup` question?

controllers/util/solr_util.go
HoustonPutman: So this should be `fsGroup` not `runAsGroup`, since `fsGroup` is meant to be used for volumes.

HoustonPutman:
> So I have a general question about this. We only specify the `fsGroup` [by default](https://github.com/apache/solr-operator/blob/main/controllers/util/solr_util.go#L479). Do we not want to use the `fsGroup` provided in the custom `PodSecurityContext`?

We do want to use the `fsGroup`, since that is what is used for volumes. I am also making sure to use `8983` as the default `fsGroup` when a custom podSecurityContext is provided that doesn't use an fsGroup.

a-nych:
Hi,

Sorry for not responding - somehow missed these notifications. I will take
a thorough look at this tomorrow.

pon., 17 kwi 2023, 22:40 użytkownik Houston Putman ***@***.***>
napisał:

HoustonPutman:
Thanks @a-nych , I'm planning on trying to cut the release by early afternoon tomorrow (EST timezone), so if you could get your review in early that'd be great!

controllers/controller_utils_test.go
```diff
@@ -906,11 +906,13 @@ var (
 	}
 	one                    = int64(1)
 	two                    = int64(2)
+	three                  = int64(2)
```
a-nych, line 909: A typo, it should be `int64(3)`.

HoustonPutman: Thanks for catching this!

PR #530, opened 2023-03-09
Add readinessCondition to stop traffic to pods that will be stopped

### LINKED ISSUE #529: Support disruption free rolling restart
As discussed in slack https://apachesolr.slack.com/archives/C022UMAPZ0V/p1676970790552379

When the operator restarts the cluster, e.g. during a version upgrade, there is no guarantee that a Solr POD is marked as not ready before `solr stop` is called. Thus clients may experience connection error during the restart. 

@HoustonPutman suggests we can implement a custom readiness gate https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate to control this better.

HoustonPutman:
Resolves #529

This uses the [pod readiness gates/conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-status) to ensure that traffic is stopped to pods that are about to be deleted for various reasons.

This will only work for those that use the managed update feature.

Things to do:
- [ ] Add Readiness Condition to podTemplate
- [ ] Make sure readinessCondition is set to `True` initially
- [ ] Set readinessCondition to `False` before pod is evicted of replicas and stopped
- [ ] Upgrade Notes and Changelog entry
- [ ] Integration Tests
- [ ] Unit Tests?

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -67,6 +67,7 @@ rules:
   - pods/status
   verbs:
   - get
+  - patch
 - apiGroups:
   - ""
   resources:
--- a/controllers/pod_lifecycle_util.go
+++ b/controllers/pod_lifecycle_util.go
@@ -1,0 +1,120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers
+
+import (
+	"context"
+	solrv1beta1 "github.com/apache/solr-operator/api/v1beta1"
+	"github.com/apache/solr-operator/controllers/util"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+func DeletePodForUpdate(ctx context.Context, r *SolrCloudReconciler, instance *solrv1beta1.SolrCloud, pod *corev1.Pod, podHasReplicas bool, logger logr.Logger) (requeueAfterDuration time.Duration, err error) {
+	// Before doing anything to the pod, make sure that users cannot send requests to the pod anymore.
+	if updatedPod, e := EnsurePodStoppedReadinessCondition(ctx, r, pod, PodUpdate, logger); e != nil {
+		err = e
+		return
+	} else {
+		pod = updatedPod
+	}
+
+	// If the pod needs to be drained of replicas (i.e. upgrading a pod with ephemeral storage), do that before deleting the pod
+	deletePod := false
+	if podHasReplicas {
+		// Only evict pods that contain replicas in the clusterState
+		if evictError, canDeletePod := util.EvictReplicasForPodIfNecessary(ctx, instance, pod, logger); evictError != nil {
+			err = evictError
+			logger.Error(err, "Error while evicting replicas on pod", "pod", pod.Name)
+		} else if canDeletePod {
+			deletePod = true
+		} else {
+			// Try again in 5 seconds if cannot delete a pod.
+			requeueAfterDuration = time.Second * 5
+		}
+	} else {
+		// If a pod has no replicas, then update it when asked to
+		deletePod = true
+	}
+
+	// Delete the pod
+	if deletePod {
+		err = r.Delete(ctx, pod, client.Preconditions{
+			UID: &pod.UID,
+		})
+		if err != nil {
+			logger.Error(err, "Error while killing solr pod for update", "pod", pod.Name)
+		}
+
+		// TODO: Create event for the CRD.
+	}
+
+	return
+}
+
+// PodStopReason describes the reason why a Pod is being stopped.
+type PodStopReason string
+
+const (
+	PodUpdate            PodStopReason = "PodUpdate"
+	StatefulSetScaleDown PodStopReason = "StatefulSetScaleDown"
+)
+
+func EnsurePodStoppedReadinessCondition(ctx context.Context, r *SolrCloudReconciler, pod *corev1.Pod, reason PodStopReason, logger logr.Logger) (updatedPod *corev1.Pod, err error) {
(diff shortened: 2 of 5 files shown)


controllers/util/solr_update_util_test.go
```diff
@@ -180,6 +205,26 @@ func TestPickPodsToUpgrade(t *testing.T) {
...
 	assert.ElementsMatch(t, []string{"foo-solrcloud-0"}, podsToUpgrade, "Incorrect set of next pods to upgrade. The overseer should be upgraded when everything is healthy and it is the last node, even though this SolrCloud resource doesn't manage all Nodes")
+
+	/*
+		Test when some pods have already been selected for deletion
+	*/
+
+	// Normal inputs
+	maxshardReplicasUnavailable = intstr.FromInt(1)
+	podsToUpgrade = getPodNames(pickPodsToUpdate(solrCloud, someScheduledForDeletionPods, testHealthyClusterStatus, overseerLeader, 6, log))
+	assert.ElementsMatch(t, []string{}, podsToUpgrade, "Incorrect set of next pods to upgrade. Do to replica placement, only the node with the least leaders can be upgraded and replicas.")
```
janhoy, line 216: *Due* to?

docs/solr-cloud/managed-updates.md
```diff
@@ -89,3 +89,25 @@ spec:
...
+
+The Solr Operator sets up at least two Services for every SolrCloud.
+- Always
+  - A clusterIP service for all solr nodes (What we call the "common service")
+- Either
+  - A [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for individual Solr Node endpoints that are not exposed via an Ingress.
+  - Individual clusterIP services for Solr Nodes that are exposed via an Ingress
+
+Only the common service uses the `publishNotReadyAddresses: false` option, since the common service should load balance between all available nodes.
+The second two services have individual endpoints for each node, so there is no reason to de-list pods that are not available.
```
janhoy, line 103:
```suggestion
The other services have individual endpoints for each node, so there is no reason to de-list pods that are not available.
```

janhoy (commented):
I'm not in a position to grok the entire PR, and I have not tested it either. But have some minor comments.

markrmiller (approved):
Not familiar with Go or these kubernetes readiness gates, but I did little reading, tried putting the changes through a couple tools, did a basic logical read through, looks like a good patch to me and a nice improvement.

controllers/util/solr_update_util_test.go
HoustonPutman:
```suggestion
	assert.ElementsMatch(t, []string{}, podsToUpgrade, "Incorrect set of next pods to upgrade. Due to replica placement, only the node with the least leaders can be upgraded and replicas.")
```

config/rbac/role.yaml
```diff
@@ -67,6 +67,7 @@ rules:
   - pods/status
   verbs:
   - get
+  - patch
```
gerlowskija, line 70: [0] Just want to make sure I understand how this stuff works:

This is a generated file, right?  Or is this one maintained by hand?

And this line was added because of the patch call made [here](https://github.com/apache/solr-operator/pull/530/files#diff-9f5886fd88f144a700c67988483eaba63bacab421a67dbf78adbd734bcbbdd00R126)?  Really cool that the generator is set up to that!

controllers/solr_pod_lifecycle_util.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	return
+}
+
+// PodConditionEquals check if a podCondition equals what is expected
+func PodConditionEquals(pod *corev1.Pod, conditionType corev1.PodConditionType, reason PodConditionChangeReason) (equal bool) {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == conditionType {
+			equal = string(reason) == condition.Reason
```
gerlowskija, line 251: [Q] Should this function 'break' or early-return after this equality test, or do actually want to continue iterating over the conditions?

docs/solr-cloud/managed-updates.md
```diff
@@ -89,3 +89,25 @@ spec:
...
+- Always
+  - A clusterIP service for all solr nodes (What we call the "common service")
+- Either
+  - A [Headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for individual Solr Node endpoints that are not exposed via an Ingress.
+  - Individual clusterIP services for Solr Nodes that are exposed via an Ingress
+
+Only the common service uses the `publishNotReadyAddresses: false` option, since the common service should load balance between all available nodes.
+The other services have individual endpoints for each node, so there is no reason to de-list pods that are not available.
+
+When doing a rolling upgrade, or taking down a pod for anytime, we want to first stop all requests to this pod.
```
gerlowskija, line 105: [0] Typo "anytime"->"any time"

docs/solr-cloud/managed-updates.md
```diff
@@ -89,3 +89,25 @@ spec:
...
+Solr will do this while stopping by first taking itself out of the cluster's set of `liveNodes` , so that other Solr nodes and clients think it is not running.
+However, for ephemeral clusters we are also evicting data before the pod is deleted. So we want to stop requests to this node since the data will soon no-longer live there.
+
+Kubernetes allows the Solr Operator to control whether a pod is considered `ready`, or available for requests, via [readinessConditions/readinessGates](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate).
+When the Solr Operator begins the shut-down procedure for a pod, it will first set a `readinessCondition` to `false`, so that no loadBalanced requests (through the common service) go to the pod.
+This readinessCondition will stay set to `false` until the pod is deleted and a new pod is created in its place.
+
+**For this reason, its a good idea to avoid very aggressive [Update Strategy](solr-cloud-crd.md#update-strategy), since a high `maxPodsUnavailable` will lead to all requests through the common service going to a small number of pods.**
```
gerlowskija, line 113: [0] Typo: "avoid very aggressive Update Strategy" -> "avoid very aggressive Update Strategies"

gerlowskija (approved):
Had a few minor comments and typo corrections, but otherwise LGTM!

config/rbac/role.yaml
HoustonPutman: This is a generated file, and it would be really cool if it was automatically added based on what commands you use... But alas its a little less cool.

This is what determines this (each controller has a separate list, that all get combined to make this file): https://github.com/apache/solr-operator/pull/530/files#diff-27cb619f80d28ac158f7885b7925d8582babfa7fe1f824599d65a745302d237cR67

docs/solr-cloud/managed-updates.md
HoustonPutman:
```suggestion
**For this reason, its a good idea to avoid very aggressive [Update Strategies](solr-cloud-crd.md#update-strategy), since a high `maxPodsUnavailable` will lead to all requests through the common service going to a small number of pods.**
```

controllers/solr_pod_lifecycle_util.go
HoustonPutman: Yeah, it should definitely break afterwards.

docs/solr-cloud/managed-updates.md
HoustonPutman: Actually going to change to "any reason", makes much more sense.

Issue #538, opened 2023-03-31
Make PodDisruptionBudget an optional feature


HoustonPutman:
The PDB was added in: https://github.com/apache/solr-operator/issues/471

Not all users are going to want to use a `PodDisruptionBudget` when running Solr.
The feature should be able to be disabled/enabled via the SolrCloud CRD.

When looking at https://github.com/apache/solr-operator/issues/471, there are other options we might want to add like PDBs per-pod.
So whatever field we add to the CRD should be expandable to customize other parts of the PDB creation/management.

PR #540, opened 2023-03-31
Make PodDisruptionBudget an optional feature

HoustonPutman:
resolves https://github.com/apache/solr-operator/issues/538

List to do:

- [x] Create CRD Options
- [x] Implement feature in controllers
- [x] Add unit tests
- [x] Add Helm values and docs
- [x] Implement Helm values in SolrCloud creation
- [x] Add docs
- [x] Add changelog


gerlowskija:
Taking a look at this now as a way to brush up on operator code.  Thanks for all the work you put into documenting this Houston!

gerlowskija:
Alright, should be ready for review.  Lmk if I missed anything!

api/v1beta1/solrcloud_types.go
```diff
@@ -758,6 +762,33 @@ type ManagedUpdateOptions struct {
...
+type SolrAvailabilityOptions struct {
+	// Define PodDisruptionBudget(s) to ensure availability of Solr
+	// +optional
+	PodDisruptionBudget SolrPodDisruptionBudgetOptions `json:"podDisruptionBudget,omitempty"`
+}
+
+type SolrPodDisruptionBudgetOptions struct {
+	// What method should be used when creating PodDisruptionBudget(s)
+	// +kubebuilder:default=true
+	Enabled bool `json:"enabled"`
```
HoustonPutman, line 774: Let's go ahead and make this a pointer. I think we should steer-clear of booleans that default to true but are not pointers (or are a child of a pointer-object, e.g. if PodDisruptionBudget was a pointer).

config/crd/bases/solr.apache.org_solrclouds.yaml
```diff
@@ -89,6 +89,27 @@ spec:
...
+                        description: What method should be used when creating PodDisruptionBudget(s)
+                        type: boolean
+                      method:
+                        default: ClusterWide
+                        description: What method should be used when creating PodDisruptionBudget(s)
+                        enum:
+                        - ClusterWide
+                        type: string
+                    required:
+                    - enabled
```
HoustonPutman, line 110: I wonder why it has `enabled` but not `method`.... How strange. Nothing for you to worry about if the tests are passing.

controllers/solrcloud_controller_test.go
```diff
@@ -88,6 +88,12 @@ var _ = FDescribe("SolrCloud controller - General", func() {
 						InitContainers:     extraContainers2,
 					},
 				},
+				Availability: solrv1beta1.SolrAvailabilityOptions{
```
HoustonPutman, line 91: Since we want this enabled by default, I think we should remove this and test that the default does its job.

Or, keep this and add the `expectPodDisruptionBudget(ctx, solrCloud, solrCloud.StatefulSetName(), statefulSet.Spec.Selector, intstr.FromString(util.DefaultMaxPodsUnavailable))` line to another test (without removing the PDB, just do the existence check). Yeah I like this option more. Keep what you have and add the PDB check to another test in this file.

controllers/solrcloud_controller_test.go
```diff
@@ -176,6 +187,12 @@ var _ = FDescribe("SolrCloud controller - General", func() {
 					},
 					RestartSchedule: "@every 30m",
 				},
+				Availability: solrv1beta1.SolrAvailabilityOptions{
+					PodDisruptionBudget: solrv1beta1.SolrPodDisruptionBudgetOptions{
+						Enabled: true,
```
HoustonPutman, line 192: For this one, set `Enabled: false`. That way we can check that it never actually creates one in the first place, and doesn't try to do a delete.

That way you can delete the two lines below:
```
expectPodDisruptionBudget(...)
expectSolrCloudWithChecks(...)
```

docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -99,8 +99,17 @@ Under `SolrCloud.Spec.updateStrategy`:
 ### Pod Disruption Budgets
 _Since v0.7.0_
 
-The Solr Operator will create a [`PodDisruptionBudget`](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) to ensure that Kubernetes does not take down more than acceptable amount of SolrCloud nodes at a time.
-The PDB's `maxUnavailable` setting is populated from the `maxPodsUnavailable` setting in `SolrCloud.Spec.updateStrategy.managed`.
+The Solr Operator can optionally create a [`PodDisruptionBudget`](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) to ensure that Kubernetes does not take down more than acceptable amount of SolrCloud nodes at a time.
```
HoustonPutman, line 102:
```suggestion
The Solr Operator can optionally create a [`PodDisruptionBudget`](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) to ensure that Kubernetes does not take down more than an acceptable amount of SolrCloud nodes at a time.
```

helm/solr/templates/solrcloud.yaml
```diff
@@ -108,6 +108,12 @@ spec:
     {{- end }}
   {{- end }}
 
+  {{- if and (.Values.availability) (.Values.availability.podDisruptionBudget) }}
```
HoustonPutman, line 111: I don't think you need this `and`... I think`.Values.availability.podDisruptionBudget` will return false if `.Values.availability` doesn't exist.

controllers/solrcloud_controller.go
```diff
@@ -464,31 +464,37 @@ func (r *SolrCloudReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-
-		// Update the found PodDistruptionBudget and write the result back if there are any changes
-		if needsUpdate && err == nil {
-			pdbLogger.Info("Updating PodDisruptionBudget")
-			err = r.Update(ctx, foundPDB)
+		if err != nil {
+			return requeueOrNot, err
+		}
+	} else { // PDB is disabled, make sure that we delete any previously created pdb that might exist.
+		err = r.Client.Delete(ctx, pdb)
```
HoustonPutman, line 494: Interesting... So it's basically as good to delete something that might not exist as it is to check if it exists first before deleting it.

HoustonPutman (commented):
Looks great!

A few hopefully quick changes, but overall great!

helm/solr/templates/solrcloud.yaml
gerlowskija: I actually tried that first - I get a helm parse error.

```
$ helm template .
Error: template: solr/templates/solrcloud.yaml:111:16: executing "solr/templates/solrcloud.yaml" at <.Values.availability.podDisruptionBudget>: nil pointer evaluating interface {}.podDisruptionBudget
```

HoustonPutman (commented):
Looks good to me!!

helm/solr/templates/solrcloud.yaml
HoustonPutman: Ahh nice, that makes sense!

PR #541, opened 2023-03-31
Use HPA v2 in tutorial

beettlle:
Updating the version of the Autoscaler as it's deprecated. 
`Warning: autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler`

And fixing the apiVersion as it's '.org' and not '.com'
```
% kubectl api-resources -o wide | grep SolrCloud
solrclouds                        solr          solr.apache.org/v1beta1                true         SolrCloud 
```


HoustonPutman (approved):

HoustonPutman:
Thanks for finding this!

PR #545, opened 2023-04-05
Patch Solr resource statuses instead of updating them

### LINKED ISSUE #544: Use patch when updating the statuses of resources.
Currently the most confusing logging for users is:

```
2023-04-05T10:27:56-04:00	ERROR	Reconciler error	{"controller": "solrcloud", "controllerGroup": "solr.apache.org", "controllerKind": "SolrCloud", "SolrCloud": {"name":"foo","namespace":"default"}, "namespace": "default", "name": "foo", "reconcileID": "87f9ba03-d059-499b-835c-bf94117edf6e", "error": "Operation cannot be fulfilled on solrclouds.solr.apache.org \"foo\": the object has been modified; please apply your changes to the latest version and try again"}
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
  	/Users/houstonputman/dev/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:329
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
  	/Users/houstonputman/dev/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:274
  sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
  	/Users/houstonputman/dev/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.2/pkg/internal/controller/controller.go:235
```

This occurs when we try to update the SolrCloud status at the end of the reconcile loop.
Since we already have the before state and the after state, it should be easy to change this from an "Update" to a "Patch", which will make this error go away.

We do need to ensure that all fields have the correct annotations so that patching works as expected.
These annotations can be found here: https://book.kubebuilder.io/reference/markers/crd-validation.html

Note: while eventually we want these annotations to be correct across the whole of the CRD, this ticket only requires that they be correct for the status SubResources.

HoustonPutman:
Fixes #544

### DIFF
--- a/api/v1beta1/solrbackup_types.go
+++ b/api/v1beta1/solrbackup_types.go
@@ -130,6 +130,7 @@ type IndividualSolrBackupStatus struct {
 	Successful *bool `json:"successful,omitempty"`
 
 	// Whether the backup has finished
+	// +optional
 	Finished bool `json:"finished,omitempty"`
 }
 
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -1071,18 +1071,26 @@ type ZookeeperPodPolicy struct {
 // SolrCloudStatus defines the observed state of SolrCloud
 type SolrCloudStatus struct {
 	// SolrNodes contain the statuses of each solr node running in this solr cloud.
+	//+listType:=map
+	//+listMapKey:=name
 	SolrNodes []SolrNodeStatus `json:"solrNodes"`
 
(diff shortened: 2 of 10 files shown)


HoustonPutman:
I went through and looked at the logs of the unit tests/integration tests and no longer saw errors related to updating statuses! Think this should be good-to-go.

PR #548, opened 2023-04-12
feat: test backup directory is writable to avoid chown

smoldenhauer-ish:


HoustonPutman:
Please add a changelog entry. You can find an example here: https://github.com/apache/solr-operator/pull/539/files#diff-5e33a9126c77070c5c0eb99dda42378496c379464576c543c5f9a731da91284f

Basically a short description, and links to the Github Issue and Github PR.

smoldenhauer-ish:
We did some (manual) tests with the changed statement. 
If the backup directory already contains several backups from ~7x400 collections the execution time of the init container cp-solr-xml dropped from >10minutes (executing the chown) down to 2-3 seconds (executing only the directory writable test)

Issue #549, opened 2023-04-12
Upgrade to the Zookeeper Operator v0.2.15


HoustonPutman:
This version was released a week ago and it would be great to get this in for the `v0.7.0` release.

https://github.com/pravega/zookeeper-operator/releases/tag/v0.2.15

PR #551, opened 2023-04-12
Remove pod labels fix after zk-op upgrade

### LINKED ISSUE #490: Zookeeper pod label customization not working
## Description

When trying to apply custom labels to Zookeeper pods through the SolrCloud spec, the labels do not show up on the final set of Zookeeper pods. 

The issue appears to be that the Zookeeper Operator is not propagating the `zookeepercluster.spec.pod.labels` fields, but only applying labels which are defined in `zookeepercluster.spec.labels`.

## Related Issues

https://github.com/apache/solr-operator/issues/148
https://github.com/pravega/zookeeper-operator/issues/511

## Possible Solutions / Workarounds

Solution: Zookeeper Operator fixes the code to ensure the `pod.labels` are applied to the final StatefulSet for the Zookeeper ensemble.

Workaround: Change the Solr Operator to create the ZookeeperCluster custom resource to apply labels to `zookeepercluster.spec.labels` instead of `zookeepercluster.spec.pod.labels`.

HoustonPutman:
Resolves #490

The fix for Zookeeper pod labels is no longer necessary since #549 has been merged in for the `v0.7.0` release. We can go back to using the native Zookeeper operator logic to set pod labels.

### DIFF
--- a/controllers/solrcloud_controller_zk_test.go
+++ b/controllers/solrcloud_controller_zk_test.go
@@ -19,7 +19,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 
 	solrv1beta1 "github.com/apache/solr-operator/api/v1beta1"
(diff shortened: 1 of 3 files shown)


HoustonPutman:
@endzyme You ok with reverting this logic and just setting the ZK resource field?

endzyme:
> @endzyme You ok with reverting this logic and just setting the ZK resource field?

Sorry for the delayed reply. Thanks for handling that!

Issue #560, opened 2023-04-24
Create a lock for involved SolrCloud operations


HoustonPutman:
The Solr Operator has started to perform pretty involved management operations, including rolling upgrades (that includes data migration for ephemeral clusters), and scaling (that can include data migration from decommissioned pods and to newly created pods).

These operations will not necessarily behave well together, so it stands to reason that we should only allow one operation at a time to occur.

Another important thing here, is that many of these operations require doing async collections API calls to Solr.
It's imperative that these operations are finished once they begin.
So without a locking mechanism, that stores the state of these operations outside of the expected state of the SolrCloud, the cloud could easily be reverted back to a previous state that no longer requires the operation.
Since the solr operator is stateless, it would not know that these operations were taking place in the previous reconcile loop, and thus a management operation could be halted halfway through.

Each management command can be locked down to ensure that this doesn't happen, however with a centralized locking mechanism, its easy to ensure that any "lockable" cluster operation must finish once it has been started.

A good place to store this metadata is in the StatefulSet annotations.
We can expect that users won't try to update these themselves, and the only time they might go away is if the StatefulSet itself is deleted.
In this case, the Solr data might no longer be there, and there is no guarantee as to the state of the cluster once the StatefulSet is re-created.
So it's probably best to stop the cluster operation anyways.

Currently the two "lockable" cluster operations would be:
- Rolling Restarts
- Scaling

HoustonPutman:
The only thing left to do here, is mitigate failures during the lockable operations.

Such as a user can provide a bad config while a scale up is happening. If that were the case, the new pods would not come up potentially, and the scale up will never unlock. (and a rolling restart can't happen, because the cluster is locked). The scale up should be able to fail, give up its lock and try again later.

PR #561, opened 2023-04-24
Add managed scale-down for SolrClouds

### LINKED ISSUE #559: Support managed scale down of SolrClouds
Parent Ticket: https://github.com/apache/solr-operator/issues/536

Currently when a SolrCloud is scaled down (the `SolrCloud.spec.replicas` option decreases), replicas are left on the Solr Pods that are being decommissioned.
This is problematic, because the cluster state will be unhealthy until the SolrCloud is scaled back up, and that pod is recreated.

When doing a rolling restart of SolrClouds with ephemeral data, the Solr operator will move data (replicas) off of a Solr pod before that Solr pod is deleted. This same logic can be used to ensure that the clusterState of Solr is healthy as the cluster is scaled down.

Right now, the safest way of ensuring this is to do a scale down, 1 pod at a time.
The current [Solr REPLACENODE API](https://solr.apache.org/guide/solr/latest/deployment-guide/cluster-node-management.html#replacenode) does not accept a list of nodes to put the new replicas on.
Therefore, if we were trying to remove the last two pods in the cluster at the same time, we couldn't ensure that the replicas of one decommissioned pod don't end up on the other decommissioned pod.

However there is an exception of the cluster is scaling down to 0 pods.
There are a couple of things we could do in this case:
- Delete all collections
- Leave the replicas there (if there is persistent storage and we are not deleting PVCs that have been scaled down)

I say that in the beginning we just use the second option, as I don't think it will be a popular thing to do anyways, and we can always add in the deletion of all data at another time.

HoustonPutman:
Resolves #559 

This also adds cluster-operation locking for SolrClouds, as described here: #560 

TODO:

- [x] Implement scale-down of SolrClouds, moving replicas when needed
- [x] Add cluster-operation locking for SolrClouds, but do not move other cluster-operations logic. That can be done separately.
- [x] Add option to do managed scale-downs, this should be enabled by default
- [x] Add integration tests for scaling down
- [x] Add documentation & upgrade notes
- [x] Add changelog entry


HoustonPutman:
Review request: @radu-gheorghe and @tflobbe. I'll finish this up today or Monday

HoustonPutman:
Cool, docs and e2e tests are pretty good to go. There might be optimizations that can be made down the line, but I think this is complete for now!

Overall the only thing that is in my mind is that we might want to change the `autoscaling` section to just be called `scaling`. Not exactly sure though. Would welcome opinions.

controllers/solr_cluster_ops_util.go
```diff
@@ -0,0 +1,218 @@
...
+				if err = r.Patch(ctx, patchedStatefulSet, client.StrategicMergeFrom(statefulSet)); err != nil {
+					logger.Error(err, "Error while patching StatefulSet to remove unneeded clusterLockOp annotation for scaling to the current amount of nodes")
+				} else {
+					statefulSet = patchedStatefulSet
+				}
+			}
+
+			// Scale down the statefulSet to represent the new number of utilizedPods, if it is lower than the current number of pods
+			// Also remove the "scalingToNodes" annotation, as that acts as a lock on the cluster, so that other operations,
+			// such as scale-up, pod updates and further scale-down cannot happen at the same time.
```
tflobbe, line 96: Isn't there a `lock` annotation being used ((as I see blow)?

controllers/util/solr_update_util.go
```diff
@@ -503,82 +503,65 @@ func GetAllManagedSolrNodeNames(solrCloud *solr.SolrCloud) map[string]bool {
...
-					if message, err = solr_api.DeleteAsyncRequest(ctx, solrCloud, requestId); err != nil {
-						logger.Error(err, "Could not delete Async request status.", "requestId", requestId, "message", message)
-					} else {
-						canDeletePod = false
-					}
+		} else {
+			logger.Info("Found async status", "requestId", requestId, "state", asyncState)
+			// Only continue to delete the pod if the ReplaceNode request is complete and successful
+			if asyncState == "completed" {
+				canDeletePod = true
```
tflobbe, line 550: Do you think it makes sense to check again for the pod before proceeding to delete it? Maybe keep doing `REPLACENODE` until it's empty? there is still a race condition, but much shorter

docs/solr-cloud/autoscaling.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Solr Pod Scale-Down
+
+When the desired number of Solr Pods that should be run `SolrCloud.Spec.Replicas` is decreased,
+the `SolrCloud.spec.autoscaling.vacatePodsOnScaleDown` option determines whether the Solr Operator should move replicas
+off of the pods that are about to be deleted.
+
+When a StatefulSet, which the Solr Operator uses to run Solr pods, has its size decreased by `x` pods, it's the last
+`x` pods that are deleted. So if a StatefulSet `tmp` has size 4, it will have pods `tmp-0`, `tmp-1`, `tmp-2` and `tmp-3`.
+If that `tmp` then is scaled down to size 2, then pods `tmp-2` and `tmp-3` will be deleted because they are `tmp`'s last pods numerically.
```
tflobbe, line 54: > then pods `tmp-2` and `tmp-3` will be deleted because they are `tmp`'s last pods numerically

May be 
> then pod `tmp-3` will be deleted first, followed by `tmp-2`because they are `tmp`'s last pods numerically

docs/solr-cloud/autoscaling.md
```diff
@@ -0,0 +1,85 @@
...
+
+If Solr has replicas placed on the pods that will be deleted as a part of the scale-down, then it has a problem.
+Solr will expect that these replicas will eventually come back online, because they are a part of the clusterState.
+However, the Solr Operator has no expectations for these replicas to come back, because the cloud has been scaled down.
+Therefore, the safest option is to move the replicas off of these pods before the scale-down operation occurs.
```
tflobbe, line 59: I think this isn't very clear. I think a simpler explanation of what the operator will do would be better, something like: 
```
The Solr Operator can update the cluster state to handle the scale-down operation by moving replicas off of the soon-to-be-deleted pods
```
Or something like that. I feel all the explanation of the difference between Solr and Solr Operator just adds confusion.

docs/solr-cloud/autoscaling.md
```diff
@@ -0,0 +1,85 @@
...
+
+If `autoscaling.vacatePodsOnScaleDown` option is not enabled, then whenever the `SolrCloud.Spec.Replicas` is decreased,
+that change will be reflected in the StatefulSet immediately.
+Pods will be deleted even if replicas live on those pods.
+
+If `autoscaling.vacatePodsOnScaleDown` option is enabled, which it is by default, then the following steps occur:
+1. Acquire a cluster-ops lock on the SolrCloud. (This means other cluster operations, such as a rolling restart, cannot occur during the scale down operation)
+1. Scale down the last pod.
+   1. Mark the pod as "notReady" so that traffic is diverted away from this pod.
```
tflobbe, line 68: Question: Is this only an ingress change? traffic that goes directly to the pod should continue to succeed, right? Otherwise, there will be update errors, replicas going DOWN, unnecessarily

tflobbe (commented):
This looks great. I mostly have questions

controllers/solr_cluster_ops_util.go
```diff
@@ -0,0 +1,218 @@
...
+	if scaleDownTo == 0 {
+		// Delete all collections & data, the user wants no data left if scaling the solrcloud down to 0
+		// This is a much different operation to deleting the SolrCloud/StatefulSet all-together
+		replicaManagementComplete, err = evictAllPods(ctx, r, instance, podList, podStoppedReadinessConditions, logger)
+	} else {
+		// Only evict the last pod, even if we are trying to scale down multiple pods.
+		// Scale down will happen one pod at a time.
+		replicaManagementComplete, err = evictSinglePod(ctx, r, instance, scaleDownTo, podList, podStoppedReadinessConditions, logger)
+	}
+	// TODO: It would be great to support a multi-node scale down when Solr supports evicting many SolrNodes at once.
```
radu-gheorghe, line 144: I doubt this is ever going to help, because conditions may change (e.g. load might increase).

Although one could argue that scaling can be done in steps. For example if you have 6 nodes and 12 shards in total, it makes sense to go from 6 to 4, skipping 5, to keep the load balanced.

Just venting :)

controllers/util/solr_update_util.go
```diff
@@ -503,82 +503,65 @@ func GetAllManagedSolrNodeNames(solrCloud *solr.SolrCloud) map[string]bool {
...
+// For scale-down operations, this can be called for pods using ephemeral or persistent data.
+func EvictReplicasForPodIfNecessary(ctx context.Context, solrCloud *solr.SolrCloud, pod *corev1.Pod, podHasReplicas bool, evictionReason string, logger logr.Logger) (err error, canDeletePod bool) {
+	logger = logger.WithValues("evictionReason", evictionReason)
+	// If the Cloud has 1 or zero pods, and this is the "-0" pod, then delete the data since we can't move it anywhere else
+	// Otherwise, move the replicas to other pods
+	if (solrCloud.Spec.Replicas == nil || *solrCloud.Spec.Replicas < 2) && strings.HasSuffix(pod.Name, "-0") {
+		queryParams := url.Values{}
+		queryParams.Add("action", "DELETENODE")
+		queryParams.Add("node", SolrNodeName(solrCloud, pod.Name))
+		// TODO: Figure out a way to do this, since DeleteNode will not delete the last replica of every type...
```
radu-gheorghe, line 516: Maybe we can go crazy and just delete all the collections from the collections API's "LIST" output?

docs/solr-cloud/autoscaling.md
HoustonPutman: This isn't actually a change, its how any operator-managed pod-stop happens since the last release. See https://github.com/apache/solr-operator/pull/530. 

But to actually answer your question, it does not affect traffic going to the headless service or individual node service. So traffic going directly to the pod will not be stopped ever. The only service that actually uses the readiness conditions is the common service (ultimately used by the common ingress endpoint).

controllers/solr_cluster_ops_util.go
HoustonPutman: Ahhh yeah, this is more because right now even if someone scales down the SolrCloud by 2, say 6 -> 4, the Solr Operator has to do it one node at a time. So it will first scale down from 6 -> 5 then 5 -> 4.

I agree that the HPA is rarely going to scale down that aggressively, but it'll just be a small improvement in the existing Solr API that we can then use to make the managed multi-node scale down operations much faster.

I agree its a small thing.

docs/solr-cloud/autoscaling.md
HoustonPutman: Ahhh understood. Will use your line to make this more concise.

controllers/util/solr_update_util.go
HoustonPutman: Yeah that's certainly fair. I would love to see a Solr feature that we can say "Don't put replicas on this node", so that we can set that flag first, and then the check wouldn't be necessary. But you are right.

If I remember correctly, that's how it worked initially. So it shouldn't be a problem to go back to that.

If we do this here, we should certainly do it the same way in the scale down operations. Currently it just waits for the command to succeed, just like here.

controllers/solr_cluster_ops_util.go
HoustonPutman: This is left over from previous implementations. Thanks for catching it.

controllers/solr_cluster_ops_util.go
```diff
@@ -0,0 +1,218 @@
...
+			reason:  ScaleDown,
+			message: "Pod is being deleted, traffic to the pod must be stopped",
+			status:  false,
+		},
+	}
+
+	if scaleDownTo == 0 {
+		// Delete all collections & data, the user wants no data left if scaling the solrcloud down to 0
+		// This is a much different operation to deleting the SolrCloud/StatefulSet all-together
+		replicaManagementComplete, err = evictAllPods(ctx, r, instance, podList, podStoppedReadinessConditions, logger)
```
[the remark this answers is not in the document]
HoustonPutman: Wonderful question. So I should make this more clear here, rather than obfuscating it.

Currently I have not actually implemented data deletion. evictAllPods is basically a NO-OP, that I left around for future improvement. If you scale down to 0, the operator just leaves the data and deletes all of the pods. This is actually the same as Deleting the StatefulSet (Though slightly different with regard to PVC lifespan). The only way the data would still live is if you are using persistent storage and using the `RetainOnDelete` option for PVCs.

The idea was that maybe we should just explicitly delete all collections before scaling to 0, since they might want to scale back up and at that point the collections in ZK will no longer exist. I didn't implement this and left it for a future option. But we can make this a blocker for the next release if you think its worth-while.

I'm also just unsure if collection deletion is what we should be doing. Scaling down nodes to me doesn't include deleting data, now that I'm thinking about it more. Either way though, this code path should be made more clear as to what it actually does, which is basically a NO-OP currently.

controllers/util/solr_update_util.go
HoustonPutman: Ahhh yes.... So this relates very closely to https://github.com/apache/solr-operator/pull/561#discussion_r1195591555

Should we be deleting the replica, or instead maybe scaling up to a second pod, then doing an update, then scaling back down. Either way it's such a small edge-case (running a single solr pod without persistent data), that I'm not sure its worth the effort. Either way we should document the edge case, though it's unrelated to this ticket. We should open another issue for this.

controllers/solr_cluster_ops_util.go
```diff
@@ -0,0 +1,218 @@
...
+		if p.Name == podName {
+			pod = &p
+			break
+		}
+	}
+
+	// TODO: Get whether the pod has replicas or not
+	podHasReplicas := true
+
+	// The pod doesn't exist, we cannot empty it
```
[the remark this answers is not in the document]
HoustonPutman: Maybe the pod never came up for some reason, maybe Kubernetes pre-emptively deleted the pod for whatever reason. But you are right, most of these would result in the pod existing but not actually running.

99.99% of the time this won't happen, but better safe than sorry.

Maybe we ensure that the pod is running before trying to evict... Because if it can't be created because of resource constraints, there's no way that the eviction is going to work...

radu-gheorghe:
Awesome! I will test it with HPA then, so I can get a jump start on that tutorial as well.

controllers/solr_cluster_ops_util.go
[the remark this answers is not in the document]
HoustonPutman: Might as well just rely on the collections API command to fail then... No need to add the logic in ourselves.

HoustonPutman:
@tflobbe thank you for the wonderful suggestions! I think I have addressed all of them, I'll merge early next week in case you want to take another look, but no worries if you don't have time.

controllers/solr_cluster_ops_util.go
```diff
@@ -210,9 +219,43 @@ func evictSinglePod(ctx context.Context, r *SolrCloudReconciler, instance *solrv
...
+			for _, shardState := range colState.Shards {
+				for replica, replicaState := range shardState.Replicas {
+					if replicaState.NodeName == podNodeName {
+						replicas = append(replicas, replica)
+					}
+				}
+			}
+		}
+	}
+	if err != nil {
```
tflobbe, line 257: Why not `else`?

tflobbe (approved):
LGTM! Thanks for explaining

controllers/solr_cluster_ops_util.go
HoustonPutman: No reason, will change.

PR #566, opened 2023-05-15
Make security context configurable.

uluzox:
In order to comply to security policies (such as from Kyverno), make the k8s `securityContext` configurable via the helm chart values.

### DIFF
--- a/helm/solr-operator/README.md
+++ b/helm/solr-operator/README.md
@@ -182,6 +182,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | serviceAccount.name | string | `""` | If `serviceAccount.create` is set to `false`, the name of an existing serviceAccount in the target namespace **must** be provided to run the Solr Operator with. This serviceAccount with be given the operator's RBAC rules. |
 | resources.limits | map[string]string |  | Provide Resource limits for the Solr Operator container |
 | resources.requests | map[string]string |  | Provide Resource requests for the Solr Operator container |
+| securityContext | map[string]string | `allowPrivilegeEscalation: false, runAsNonRoot: true` | Provide security context for the Solr Operator container |
 | labels | map[string]string |  | Custom labels to add to the Solr Operator pod |
 | annotations | map[string]string |  | Custom annotations to add to the Solr Operator pod |
 | nodeSelector | map[string]string |  | Add a node selector for the Solr Operator pod, to specify where it can be scheduled |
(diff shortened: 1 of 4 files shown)


helm/solr-operator/README.md
```diff
@@ -182,6 +182,7 @@ The command removes all the Kubernetes components associated with the chart and
 | serviceAccount.name | string | `""` | If `serviceAccount.create` is set to `false`, the name of an existing serviceAccount in the target namespace **must** be provided to run the Solr Operator with. This serviceAccount with be given the operator's RBAC rules. |
 | resources.limits | map[string]string |  | Provide Resource limits for the Solr Operator container |
 | resources.requests | map[string]string |  | Provide Resource requests for the Solr Operator container |
+| securityContext | map[string]string | `allowPrivilegeEscalation: false, runAsNonRoot: true` | Provide security context for the Solr Operator container |
```
HoustonPutman, line 185:
```suggestion
| securityContext | object | `allowPrivilegeEscalation: false, runAsNonRoot: true` | Provide security context for the Solr Operator container |
```

the securityContext is an object, that can have values that aren't strings. So we shouldn't use `map[string]string` as the type.

HoustonPutman (changes requested):
This looks great, just one suggestion for the documentation. Thanks for contributing!

Would you mind also adding a changelog entry to the Chart.yaml? You will have to remove the two entries that are there as placeholders now, but you can see an example of this here: https://github.com/apache/solr-operator/commit/0539494189c6634b6ecb91ac5e09809df4c5b630#diff-5e33a9126c77070c5c0eb99dda42378496c379464576c543c5f9a731da91284f

uluzox:
> This looks great, just one suggestion for the documentation. Thanks for contributing!
> 
> Would you mind also adding a changelog entry to the Chart.yaml? You will have to remove the two entries that are there as placeholders now, but you can see an example of this here: [0539494#diff-5e33a9126c77070c5c0eb99dda42378496c379464576c543c5f9a731da91284f](https://github.com/apache/solr-operator/commit/0539494189c6634b6ecb91ac5e09809df4c5b630#diff-5e33a9126c77070c5c0eb99dda42378496c379464576c543c5f9a731da91284f)

Thanks for the review, @HoustonPutman. Could you please have another look?

HoustonPutman:
Thanks @uluzox!

PR #571, opened 2023-05-24
Add SessionAffinity and SessionAffinityConfig fields to ServiceOptions

### LINKED ISSUE #535: Allow sessionAffinity for the SolrCloud CommonService
Kubernetes ClusterIP Services allow for sessionAffinity, and customization of the max session stickiness timeout.

More information can be found here: https://kubernetes.io/docs/reference/networking/virtual-ips/#session-affinity

All we would need is these options to be given under `SolrCloud.Spec.customSolrKubeOptions.commonServiceOptions`, and then passed to the service. This doesn't make sense for the headless service or node services.

trevorpburke:
Attempts to address: https://github.com/apache/solr-operator/issues/535


api/v1beta1/common_types.go
```diff
@@ -164,6 +164,14 @@ type ServiceOptions struct {
 	// Labels to be added for the Service.
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
+
+	// Supports "ClientIP" and "None". Used to maintain session affinity. Enable client IP based session affinity. Must be ClientIP or None. Defaults to None
+	// +optional
```
HoustonPutman, line 169:
```suggestion
	// +optional
	// +kubebuilder:default=None
	// +kubebuilder:validation:Enum=None,ClientIP
```

So we can actually put the default, and the enum constraints into the CRD itself using [Kubebuilder CRD markers](https://book.kubebuilder.io/reference/markers/crd-validation.html)!

api/v1beta1/common_types.go
```diff
@@ -164,6 +164,14 @@ type ServiceOptions struct {
...
+	SessionAffinity string `json:"sessionAffinity,omitempty"`
+
+	// sessionAffinityConfig contains the configurations of session affinity.
+	// +optional
+	SessionAffinityConfig corev1.SessionAffinityConfig `json:"sessionAffinityConfig,omitempty"`
```
HoustonPutman, line 174:
```suggestion
	SessionAffinityConfig *corev1.SessionAffinityConfig `json:"sessionAffinityConfig,omitempty"`
```

Making this field a pointer, makes it easier for us to know whether the user provided one or not!

HoustonPutman (commented):
This is a great start!

So what you have done is make the configuration available in the CRD. So the user can now provide it, however it doesn't actually set the configuration on the underlying Services that are created for the SolrCloud/PrometheusExporter.

So we still need to:

- [ ] add the controller logic to set these fields when creating services
- [ ] make sure the fields are copied correctly when updating the services
- [ ] add tests to ensure the fields are propagated correctly
- [x] Add helm chart options

If you want an example of a PR that does something very similar, you can look here: https://github.com/apache/solr-operator/pull/350

I'm happy to contribute too, I just don't want to step on your toes! Let me know how I can help 🙂

api/v1beta1/common_types.go
trevorpburke: Ah, neat code-gen

api/v1beta1/common_types.go
trevorpburke: Makes sense

trevorpburke:
> This is a great start!
> 
> So what you have done is make the configuration available in the CRD. So the user can now provide it, however it doesn't actually set the configuration on the underlying Services that are created for the SolrCloud/PrometheusExporter.
> 
> So we still need to:
> 
> * [ ]  add the controller logic to set these fields when creating services
> * [ ]  make sure the fields are copied correctly when updating the services
> * [ ]  add tests to ensure the fields are propagated correctly
> * [ ]  Add helm chart options
> 
> If you want an example of a PR that does something very similar, you can look here: #350
> 
> I'm happy to contribute too, I just don't want to step on your toes! Let me know how I can help 🙂

Thanks for the feedback. I think I addressed updating the `helm/` files that were changed in the PR you linked. Can you share which of those files in #350 are generated vs manually written? 

AFAICT it looks like I need to also update `controllers/util/solr_util.go`, `controllers/util/prometheus_exporter_util.go`, and `controllers/util/common.go`.

api/v1beta1/common_types.go
```diff
@@ -169,9 +169,10 @@ type ServiceOptions struct {
 	// +optional
 	SessionAffinity string `json:"sessionAffinity,omitempty"`
 
-	// sessionAffinityConfig contains the configurations of session affinity.
 	// +optional
-	SessionAffinityConfig corev1.SessionAffinityConfig `json:"sessionAffinityConfig,omitempty"`
+	// +kubebuilder:default=None
```
HoustonPutman, line 173: These (the default and enum lines) were meant to be put up above with `sessionAffinity`, not `sessionAffinityConfig`.

helm/solr/README.md
```diff
@@ -300,6 +300,8 @@ When using the helm chart, omit `customSolrKubeOptions.`
 | statefulSetOptions.podManagementPolicy | string | `"Parallel"` | Policy for how Solr pods should be managed in the statefulSet, ["OrderedReady" or "Parallel"](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies) |
 | commonServiceOptions.annotations | map[string]string |  | Custom annotations to add to the Solr common service |
 | commonServiceOptions.labels | map[string]string |  | Custom labels to add to the Solr common service |
+| commonServiceOptions.sessionAffinity | string |  | Used to maintain session affinity. Enable client IP based session affinity |
```
HoustonPutman, line 303:
```suggestion
| commonServiceOptions.sessionAffinity | string | `"None"` | Choose session affinity to setup for the common service. Available options are "None" and "ClientIP". |
```

HoustonPutman (changes requested):
The changes look good so far! Just a few comments.

Now for actually implementing and testing!

There are three places to implement this logic.

In `controllers/util/solr_util.go`:`GenerateCommonService()`. That is what generates the common service for the solrCloud, so that's where we need to take these options from the SolrCloud and pass it to the Service.

In `controllers/util/prometheus_exporter_util.go`:`GenerateSolrMetricsService()`. This should be exactly (or almost exactly) the same logic as used above. We just need to support the same thing for the prometheus exporter, since it's a shared option amongst all services.

In `controllers/util/common.go`:`CopyServiceFields()`. This is where we update the existing service with what our "expected state" of the service is. So we are copying from the `from` service to the `to` service. You can basically copy and paste those if statements, using new ones for `sessionAffinity` and `sessionAffinityConfig`.

Then our last step will be testing!

HoustonPutman:
Also just a thought, but we are giving users the option for the headlessService and podService, but those really don't need sessionAffinity, because they are meant to send requests to a single pod, so affinity doesn't matter. So I'm fine leaving those undocumented and unused.

HoustonPutman:
I went ahead and added the rest of the tasks, with help from Claude. I think this should be good to go for the next release.

PR #576, opened 2023-06-01
Document how a simple load balancer can be achieved through annotations on the commonServiceOptions

elangelo:


docs/solr-cloud/solr-cloud-crd.md
```diff
@@ -151,6 +151,16 @@ If both of those criteria are met, then an individual ClusterIP Service will be
...
 Under this property, you can set custom `annotations`, `labels` and an `ingressClassName`.
 
+**Note:** If you have no need for an `Ingress` or a specific hostname to address your SolrCloud cluster you can create a simple loadbalancer that can be addressed from outside the kubernetes cluster. To achieve this you can add annotations to `SolrCloud.spec.customSolrKubeOptions.headlessServiceOptions.annotations`. Snippet below shows you can create an NLB on AWS:
+
+```yaml
+headlessServiceOptions:
+  annotations: 
+    service.beta.kubernetes.io/aws-load-balancer-type: external
+    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
+    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
```
HoustonPutman, line 161:
```suggestion
spec:
  customSolrKubeOptions:
    headlessServiceOptions:
      annotations: 
        service.beta.kubernetes.io/aws-load-balancer-type: external
        service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
        service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing  
```

HoustonPutman (commented):
Thank you so much for the docs, I think they would be a great addition.

One question: Is there a reason that you are using the headless service instead of the common service? Since the common service is already a clusterIP service, I think it would be more suited to this.

docs/solr-cloud/solr-cloud-crd.md
elangelo: I must admit I blindly followed what I used to do with Elasticsearch, exposing the headless service. I'll do some reading to understand what would be advantages and disadvantages

elangelo: Advantage of using the headless service: less abstraction, access to the individual pods
Disadvantage: no load balancing on the server side so client  library needs to implement load balancing themselves.

So using a headless service for my previous Elasticsearch project made sense as the client library we were using actually supported 'load balancing between nodes' through round robin.
I'm not aware of all the client libraries that we have for SolrCloud but at least in our case (just http calls through a python abstraction) it looks like indeed i should be using the common service instead.

HoustonPutman: oh interesting, I didn't realize that these annotations would add an IP per pod, I thought it would be one IP for the service. If it exposes individual Solr addresses, then that is certainly a good reason to use the headless service.

HoustonPutman (approved):
Thanks so much @elangelo , great documentation addition!

PR #610, opened 2023-08-24
Support TLS with recent Solr 9.x versions

### LINKED ISSUE #609: Fix Solr probes w/TLS for Solr 9.2+
Currently various versions of Solr 9 cannot be run with TLS in the Solr Operator, because of the way that secure probes are run. Namely, we explicitly invoke SolrCLI, which has changed its location in newer versions. We need to use the `solr api -` command, which will be available in Solr 8.11, so this can only be done when that is the minimum supported version.

HoustonPutman:
Resolves #609

This is a WIP

There are various things that Solr 9.4+ will help with, but we need to support 8.11 at the very least.

Therefore, this is a part of #616

Basically we need to:

- [x] Make sure the hostname for the Solr API is always correct
- [x] Use the `solr api` command instead of invoking `SolrCLI` explicitly
- [x] Mention SOLR_TOOL_OPTS in the future Solr version usages, and find a workaround for 8.11/9.3
- [x] Changelog entry
- [x] Upgrade Notes
- [x] Supported Versions matrix


HoustonPutman:
Thanks to [SOLR-16964](https://issues.apache.org/jira/browse/SOLR-16964), Solr 9.4 will come with the correct SNI hostCheck default given the value of CheckPeerName. Because of this, and the fact that we only really will support SSL for 9.4+, we don't need to set this value. (And it isn't an issue in 8.11, so 8.11 and 9.4+ will work without this setting).

sarkaramrit2:
Thank you @HoustonPutman

PR #611, opened 2023-08-24
Add integration tests for TLS Solr

### LINKED ISSUE #605: Add integration tests for TLS
Currently we have no way of detecting if TLS works for the supported versions of Solr.

As a part of this ticket, we should document which versions of Solr are supported with SSL in the Solr Operator, as there are some issues with Solr 9 SSL incompatibilities.

HoustonPutman:
Resolves #605

The tests will use Cert Manager (Using both the Cert Manager CSI Driver, as well as secrets).

Tests for:

- [x] Client Auth Need - Doesn't work with the CSI Driver, as the CSI Driver does not create truststores?
- [x] Client Auth Want
- [x] VerifyHostName


HoustonPutman:
The integration tests will only work with 8.11, until #610 is merged.

PR #625, opened 2023-09-20
Call rebalance API once all pods are updated

### LINKED ISSUE #615: Rebalance ephemeral Solr cloud after a rolling restart
When the solr operator does a rolling restart of a SolrCloud that uses ephemeral data, it will move replicas off of a node before restarting it. Therefore by the end of the rolling restart, at least one node won't have any replicas on it because no other nodes have had their replicas moved after it was restarted (because it was the last node to be restarted). This results in an unbalanced cluster.

The solution is to use the new Balance Replicas feature of Solr (introduced in 9.3), like we do for Scale Up operations, after all nodes are successfully live after being restarted. Now that we have locked cluster operations, this should be much easier.

radu-gheorghe:
Timid attempt to fix #615 

It works on my machine, it seems to fix the problem, but:
* I have no tests. Yet. I have a feeling this is the hard part 🙂 
* I guess I'll need to update [managed-updates.md](https://github.com/apache/solr-operator/blob/main/docs/solr-cloud/managed-updates.md) a bit as well.
* There are a couple of TODOes in my patch where I'm not sure if things are right - any feedback there would be great!
* Last but not least, with or without my patch, I always got a DOWN shard at the end of my rolling update. That's a separate issue, right? Still, it's always reproducible in my environment, maybe due to the fact that I'm testing with some docs in my collection and I'm having my IDE (with Solr Operator) talk to Kubernetes via a port forward to the "common" service, which sometimes fails because the underlying pod gets restarted, so I get a few "connection refused" errors until I put it back. Still, it should be something that the Operator handles gracefully, IMO.

Any feedback is welcome, I'd be glad to push this forward!


HoustonPutman (changes requested):
This is great! I have two edge cases that I think we need to account for:

1. In the case that all the pods are updates, the async rebalance command will be started. The user can then change the spec again, and the updatedPodCount will then again be `0`, because all pods will be out of date. At that point we will have forgotten about the rebalance command that we started. We need to make sure that we complete any async command before we start a new one, so I suggest that we split off the async command to a separate ClusterOp, that we start after the restart is complete. That way another restart will only begin once the rebalance is finished.
2. If I'm correct, we are rebalancing even if replicas weren't moved around during the rolling restart. In my opinion, we should only rebalance if replicas were moved around.

HoustonPutman (approved):
I think this is good to go @radu-gheorghe , give it a look!

radu-gheorghe:
Looks good to me, too, thanks a lot @HoustonPutman!

I already had a look yesterday, too, and noticed you couldn't help yourself and fixed your 2) above 🙂

HoustonPutman:
> I already had a look yesterday, too, and noticed you couldn't help yourself and fixed your 2) above 🙂

Haha I remembered that we had the lovely readinessConditions that told us if the cloud was ephemeral, so it was very easy to do it!

Issue #616, opened 2023-09-01
Bump minimum supported Solr version to 8.11


gerlowskija:
According to our [compatibility matrix](https://github.com/apache/solr-operator/blob/main/docs/upgrade-notes.md#solr-versions), Solr Operator 0.7.0 supports a huge range of Solr versions - stretching from Solr 9.3 all the way back to 7.7.

There's no need to support such a wide range.  And in fact there's a good bit of downside.  It dilutes our testing, and prevents the operator from relying on any Solr feature introduced in the last 2 or 3 years.  In fact, we've accumulated [quite a number of features we'd like to use](https://github.com/apache/solr-operator/blob/main/dev-docs/changes-for-minimum-solr-version.md) once our minimum version increases a bit.

It seems like a big jump, but 8.11 has been out for over a year and a half now (!).  Maybe we can target that as the minimum Solr version for operator 0.8.0.

If there's no objection to that plan, lets use this ticket to make the required changes to our docs.  And maybe as a parent/umbrella ticket for incorporating use of some of those Solr features we've been prevented from using up to this point.

gerlowskija:
We've made pretty good progress on this via the linked issues.  There's one last change we can do (involving our ZK-bootstrapping) to make sure we're taking advantage of all that 8.11 offers, and then we can close this ticket out!

HoustonPutman:
Thanks for all of the awesome work here Jason! Yeah one more and we should be good to go! (And the ZK Bootstrapping will be great to get rid of!!)

I think after that we should be good for the v0.8.0 release finally!

HoustonPutman:
Unfortunately the Chroot initialization works for 8.11, but the zkCli commands break when the chroot is not pre-initialized ([SOLR-17019](https://issues.apache.org/jira/browse/SOLR-17019)). This will be fixed in 9.4 or 9.5, but there's no real reason to do that part until those issues are addressed.

So I think this ticket can be considered closed now!

Issue #504, opened 2022-12-13
Probes for readiness and liveness should be different


janhoy:
We are using kyverno to enforce certain common k8s policies, and it complains about several things in Solr-operator.

One is this one https://kyverno.io/policies/other/ensure_probes_different/ensure_probes_different/

We have auth enabled, so it uses the alternative shell command, which is the same for both probes. Is this on purpose? Could we use different commands?

HoustonPutman:
We could use different commands. We can probably start mandating Solr 8 soon, and at that point we can use the solr healthcheck handler as the readiness check, and keep the liveness check as is. (Don't necessarily want to restart Solr because of an issue with ZK...) But if you have other suggestions, I am all ears!

HoustonPutman:
> So the liveness probe failing will cause k8s to kill the POD, and the readiness probe failing will cause traffic to be temporarily routed to other replicas, is that about right?

> Also, if you have 500 cores on a server, and only one is recovering, it would be a pity if the pod was flagged as not-ready, since Solr is capable of routing traffic to all the other cores. But this is perhaps where the PDB comes in...

Yes, but the readiness probe only affects Services that have the `PublishNotReadyAddresses` option set to `true`. Our common service (one endpoint for all nodes) has this set to true, while the headless service has this set to false. Therefore Solr can still route traffic to the example node you mentioned, as is necessary for things like recovery (All internal requests will go to node-specific endpoints, which are managed by the headless service). However users initial requests will not end up on that node if they are using the common service.

I don't really have a strong feeling either way on this, whether its a good thing or a bad thing. But if we use the healthcheck endpoint and just use it for the zk connection then that is safer. We definitely don't want to route requests to nodes that can't talk to ZK, when there are other nodes available. (Solr will deal with this itself via live_nodes hopefully).

> Also, a rolling restart uses the readiness probe as a sign that it can move on to take down the next one? So for that reason we'd like all cores to be up.

The readiness probe is also used for this, but luckily for us the ManagedUpdate option for Solr really doesn't use it a whole lot. Instead it reads the cluster state manually to see when things are healthy enough to move on to the next node(s). So the vast majority of people shouldn't be affected to much by this aspect. (One note, the readiness probe is used when talking about the number of "down" pods, not replicas, so the operator will wait till all nodes are "ready" to do the last pod restart, the overseer.)

[...]

PR #629, opened 2023-09-21
Use Solr's /admin/info/health for pod readiness checks

gerlowskija:
Now that the operator expects Solr versions >=8.11, we can use Solr's /admin/info/health endpoint as our default readiness probes.  This change also means that the operator now defaults to different Solr endpoints for liveness vs readiness probes, which satisfies what some consider a best practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Please visit the following pages for documentation on using and developing the S
   - The released helm charts and their instructions should be used for all safe and stable deployments.
     The charts found in `helm/` are not guaranteed to be compatible with the last stable release, and should only be used for development purposes.
 - [Running the Solr Operator](https://apache.github.io/solr-operator/docs/running-the-operator)
+- [Known Issues](https://apache.github.io/solr-operator/docs/known-issues)
 - Available Solr Resources
     - [Solr Clouds](https://apache.github.io/solr-operator/docs/solr-cloud)
     - [Solr Backups](https://apache.github.io/solr-operator/docs/solr-backup)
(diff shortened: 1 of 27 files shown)


HoustonPutman:
Note, this is a part of #616 

Also I think it would be good to keep the `/system` handler in use for the Liveness probe (since that says whether Solr is running or not), and use the `/health` handler for the Readiness probe. Some kubernetes environments like these probes to be different, and also I see a problem where Solr is restarted on GC pauses (causing ZK live node issues), when that would only make the problem worse. See #504 for some more information.

gerlowskija:
Ah ok, makes sense.  I didn't realize there'd been some discussion around differentiating them - will take a second pass for that shortly.

HoustonPutman (approved):
This looks great! I think one last thing is to add something in the upgrade notes that custom security.json files will need both of these probes by default starting in v0.8.0!

gerlowskija:
Ah, good call - done!

gerlowskija:
Thinking on this a bit more, are we sure we want to use `/admin/info/health` for readiness probes?

Unlike the `/system`, `/health` will return an error on nodes with bad ZK connections and prevent them from receiving traffic if ZK goes down.  That makes sense for lots of traffic like admin requests or updates - no point sending any of those requests to a SolrCloud node that has no ZK.  But it'll block other traffic that would've succeeded too - like queries, which Solr can sometimes continue to serve even without ZK.

Overall I think the additional intelligence and fidelity that `/health` gives us as a readiness probe is worth this slight degradation in how we'd route traffic in response to a total catastrophe.  Just wanted to call it out in case others disagree.  I'll add a sentence or two documenting this aspect of things and then merge, but I'll leave this out here for a few more days to give folks a chance to chime in.

PR #631, opened 2023-09-21
Add maxBooleanClauses setting to default solr.xml

gerlowskija:
This allows us to take advantage of the additional "safety-valve" setting added in SOLR-13336 back in Solr 8.1!


HoustonPutman (approved):
Perfect!

PR #633, opened 2023-09-27
Upgrade dependencies

HoustonPutman:
This includes `controller-runtime` and the Kubernetes APIs.


HoustonPutman:
Need to make sure the multi-namespace watch option still works before merging, but otherwise this should be good to go.

HoustonPutman:
Ok this should be good to go, I tested the custom namespace watching manually and it worked! Hopefully we can get that integration tested eventually, but it'd add a lot of time to the integration tests for little gain currently...

PR #634, opened 2023-09-27
Get correct SolrPods in SolrCloud controller

HoustonPutman:
The SolrCloud controller needs to fetch the list of Solr Pods for a given SolrCloud, in order to determine what to restart and construct the SolrCloud.Status object. However, when an old SolrCloud has been deleted and a new SolrCloud has been created (with the same name), there may be time when the new SolrCloud has been created, but the Pods for the old SolrCloud have not yet been deleted. So the SolrCloud controller will then think that the old Pods are related to the new SolrCloud, because they have the right labels. This isn't the case however, so we need to make sure that only the SolrCloud Pods that are actually related to that SolrCloud are fetched.

This is unlikely to show up in a production setting, but it causes havoc with the integration tests more than occasionally.

I have also added the StatefulSet json to the output for integration tests, allowing us to better debug failures.


gerlowskija (approved):
LGTM

PR #636, opened 2023-09-29
Add sysprop driven XML tags to default solr.xml

### LINKED ISSUE #635: Bring solr.xml into line with current Solr default settings.
The solr.xml that the operator currently uses as a default for new `solrcloud` resources was crafted in the Solr 7.x timeline.  [A recent issue](https://github.com/apache/solr-operator/issues/630) took some baby steps towards modernizing it, but it's still missing a number of settings that we can take advantage of now that our minimum Solr version is 8.11.  Particularly, sysprop-driven XML tags for "sharedLibs", "allowPaths", and "metrics".

Let's add these to our default solr.xml, so that users can drive these settings as desired through system properties.

gerlowskija:
The default solr.xml now has the necessary plumbing to obey the commonly used `solr.sharedLib`, `solr.allowPaths`, and `enableMetrics` system properties.

Resolves #635

### DIFF
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -61,8 +61,9 @@ const (
 
 	DefaultStatefulSetPodManagementPolicy = appsv1.ParallelPodManagement
 
-	DistLibs    = "/opt/solr/dist"
-	ContribLibs = "/opt/solr/contrib/%s/lib"
+	DistLibs              = "/opt/solr/dist"
+	ContribLibs           = "/opt/solr/contrib/%s/lib"
+	SysPropLibPlaceholder = "${solr.sharedLib:}"
 )
 
 var (
@@ -786,6 +787,8 @@ const DefaultSolrXML = `<?xml version="1.0" encoding="UTF-8" ?>
     <int name="connTimeout">${connTimeout:60000}</int>
   </shardHandlerFactory>
   <int name="maxBooleanClauses">${solr.max.booleanClauses:1024}</int>
+  <str name="allowPaths">${solr.allowPaths:}</str>
+  <metrics enabled="${metricsEnabled:true}"/>
   %s
 </solr>
 `
@@ -831,6 +834,9 @@ func GenerateSolrXMLString(backupSection string, solrModules []string, additiona
 func GenerateAdditionalLibXMLPart(solrModules []string, additionalLibs []string) string {
 	libs := make(map[string]bool, 0)
 
+	// Placeholder for users to specify libs via sysprop
+	libs[SysPropLibPlaceholder] = true
+
 	// Add all module library locations
 	if len(solrModules) > 0 {
 		libs[DistLibs] = true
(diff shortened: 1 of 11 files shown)


HoustonPutman (commented):
This looks good to me, I'm very excited to use the SOLR_MODULES option in 9.0. But until then this is great!

One question. If the user doesn't provide `solr.sharedLib`, the xml section will start with `,`. Is that ok?

HoustonPutman:
Oh, might as well include the `solr.port.advertise` here as well, unless you wanted to do that in a separate PR. No worries either way.

gerlowskija:
> If the user doesn't provide solr.sharedLib, the xml section will start with ,.

Hmm, not quite sure what you mean.  Are you talking about the [GenerateAdditionalLibXMLPart change here](https://github.com/apache/solr-operator/pull/636/files#diff-653faf42eabedf3285e433f247c993282f035ee70781d151f8c8d68fee2621a3R858)?

If so, AFAICT, 'libList' will never be empty because it at least always has `${solr.sharedLib:}`.  So if the user doesn't use any backups or modules, the value produced XML should look like:

```
<str name="sharedLib">${solr.sharedLib:}</str>
```

But, maybe I'm missing something @HoustonPutman ?

HoustonPutman:
> > If the user doesn't provide solr.sharedLib, the xml section will start with ,.
> 
> Hmm, not quite sure what you mean. Are you talking about the [GenerateAdditionalLibXMLPart change here](https://github.com/apache/solr-operator/pull/636/files#diff-653faf42eabedf3285e433f247c993282f035ee70781d151f8c8d68fee2621a3R858)?
> 
> If so, AFAICT, 'libList' will never be empty because it at least always has `${solr.sharedLib:}`. So if the user doesn't use any backups or modules, the produced XML should look like:
> 
> ```
> <str name="sharedLib">${solr.sharedLib:}</str>
> ```
> 
> But, maybe I'm missing something @HoustonPutman ?

So my thought was if `${solr.sharedLib:}` equates to an empty string, which it will by default, that Solr wouldn't handle the `,` in the beginning very well. But I just tested it out, and it works fine! So no issue there 🙂

gerlowskija:
Ah, yes - sorry, I was being dense.  Had an epiphany about what you meant while I was afk getting dinner.  I tested it as well - looks good on 8.11 and 9.3 both. So I think we're safe 😅

gerlowskija:
> might as well include the solr.port.advertise here as well

I've added the necessary changes to work `solr.port.advertise` into generated solr.xml files, though I think it'll need tweaked a bit.

The generated solr.xml was already set up to reference a different system property called `hostPort`.  It looks like we made pretty regular use of that sysprop by default: even going so far as to error out of reconciliation whenever a custom solr.xml didn't include `<int name="hostPort">${hostPort:80}</int>`.  This PR (right now) updates this all to use `solr.port.advertise` instead.

Still doing a bit of testing on this and thinking through the upgrade ramifications (little worried about what this change means for administrators with custom solr.xml files that want to upgrade their operator), but this should all be ready for review.

controllers/util/solr_util.go
```diff
@@ -134,7 +135,7 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 
 	// Keep track of the SolrOpts that the Solr Operator needs to set
 	// These will be added to the SolrOpts given by the user.
-	allSolrOpts := []string{"-DhostPort=$(SOLR_NODE_PORT)"}
```
HoustonPutman, line 137: So we can actually just use an envVar, `SOLR_PORT_ADVERTISE`, instead of relying on SOLR_OPTS. Makes the code a small bit cleaner 🙂 

But we can keep the `hostPort` in place and deprecate it (So we would be using `-DhostPort` and `SOLR_PORT_ADVERTISE` at the same time), to give people with a custom `solr.xml` a release to change their Solr.xml to use `solr.port.advertise`, like is used in the official `solr.xml`. As long as we document it in the upgrade notes, we should be good to go!

gerlowskija: Ah, right!  I forgot `bin/solr` was set up to convert `SOLR_PORT_ADVERTISE` to a sysprop that solr.xml can take advantage of, cool.  👍

HoustonPutman (approved):
One little suggestion for the upgrade notes, but otherwise it looks good to me!

PR #639, opened 2023-10-02
Pass exporter args as env vars where possible

### LINKED ISSUE #638: Replace exporter CLI args with env vars where possible
In recent versions of Solr (>=8.8) the prometheus exporter has an env-var equivalent for each CLI argument taken in by its start script.

Env vars can be easier to read and parse than a long string of CLI options - let's switch the operator over to using env-vars to pass certain prometheusexporter.Spec values, instead of building an array/string of arguments.

(This is a part of #616)

gerlowskija:
In recent versions of Solr (>=8.8) the prometheus exporter has an env-var equivalent for each CLI argument taken by its start script.

This commit switches the operator over to providing values using this env-var capability where possible.  Hopefully this makes the created resources a little cleaner and easier-to-read for administrators and users.

Closes #638

### DIFF
--- a/controllers/solrprometheusexporter_controller_test.go
+++ b/controllers/solrprometheusexporter_controller_test.go
@@ -103,14 +103,29 @@ var _ = FDescribe("SolrPrometheusExporter controller - General", func() {
 			Expect(deployment.Spec.Template.Spec.Containers).To(HaveLen(1), "Wrong number of containers for the Deployment")
 			Expect(deployment.Spec.Template.Spec.InitContainers).To(BeEmpty(), "Wrong number of initContainers for the Deployment")
 
-			expectedArgs := []string{
-				"-p", strconv.Itoa(util.SolrMetricsPort),
-				"-n", strconv.Itoa(int(testNumThreads)),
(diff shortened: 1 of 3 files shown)


gerlowskija:
Two questions I could use a bit of clarity on before this can be merged:

1. How do we want to handle the potential conflict between user-specified env-vars, and values like `.Spec.NumThreads` that are converted into an env-var by the operator?  Currently in this PR, the user-specified env-var "wins" that conflict, though I'm not sure that's the best behavior.
2. Does this need a "changes" helm chart entry, or can that be skipped here as this is largely a refactor?

HoustonPutman:
> 1. How do we want to handle the potential conflict between user-specified env-vars, and values like `.Spec.NumThreads` that are converted into an env-var by the operator?  Currently in this PR, the user-specified env-var "wins" that conflict, though I'm not sure that's the best behavior.

I think that user-specified env-vars should win.

> 2. Does this need a "changes" helm chart entry, or can that be skipped here as this is largely a refactor?

Yeah I guess it doesn't need it, could go either way, so I don't particularly care!

HoustonPutman (approved):
Great work!

Issue #624, opened 2023-09-20
Possibility to run as readOnlyRootFilesysystem


thomaswoeckinger:
In Openshift environments (may in others too) it is possible to restrict containers with SecurityContextConstraints (SCC).

Especial setting the root filesystem to read only would increase security.

Currently the init container `setup-zk`? is preventing read only root filesystem, as it is writing  to `/tmp`.

So if this container would mount an `emptyDir` to /tmp it would be possible to set the root filesystem to read only.

thomaswoeckinger:
It is not that complicated, it is sufficient to use an emptyDir and mount it to /tmp. This is because readOnlyRootFilesystem is not used wirh mount points.

HoustonPutman:
Yeah @radu-gheorghe , Thomas is not saying that we need to be able to specify `readOnlyFilesystem`, as that is already possible. We just need to make sure any folder that we write to in Solr or in the init containers is backed by a volume (ephemeral volumes by default), so that we don't see an error when the `readOnlyFilesystem` option is used.

radu-gheorghe:
OK, so I'll change `zk-init` to mount an `emptyDir` into `/tmp`. I'll also add `readOnlyFilesystem` to its definition in order to test it, but I'd like to leave it like that, I don't see a reason not to. Sounds good?

And I'll also try to test with `readOnlyFilesystem` everywhere. Maybe the Solr container also writes to `/tmp` or somewhere funky. I couldn't do that so far, but I'll press on 😁

PR #648, opened 2023-10-26
Support readOnlyRootFilesystem

radu-gheorghe:
This should be the first part (or at least an initial attempt at it) for #624 

It adds /tmp as emptyDir to the `zk-init` init container. It also sets `readOnlyRootFilesystem` to it. For testing and also why not 🙂 I've tested it with Solr 9.3 and a chroot and I see the chroot and I see `readOnlyRootFilesystem` in its container definition.

Any feedback is welcome. Once we agree about this, I'd like to check if Solr itself can run as `readOnlyRootFilesystem` and if not, add support for that, check for issues and iron them out.


radu-gheorghe:
I just pushed a commit that should fix the breaking tests.

I've also experimented with the Solr container using `readOnlyRootFilesystem` by doing `kubectl edit` on the StatefulSet. Solr seems to work just fine (I created a collection, added some docs...). Nothing seems to break, at least at first. Maybe this fixes your problem in #624 @thomaswoeckinger ?

thomaswoeckinger:
I think so, i will test it on my side and give you feedback then.

thomaswoeckinger:
From my point of view, at least some common request for collection creation, modification, deletion and sharding is working, so i think this is good to go

HoustonPutman:
Let's keep it just the zk-init container for now (or all the init containers we do). That way we don't break use cases we don't know about. Not sure there's a giant benefit for us to set it by default right now.

radu-gheorghe:
Sounds good to me! Does this mean that this PR is good to merge, or should I change something?

HoustonPutman:
I added a changelog entry. Once the tests pass, I'll merge. Thanks for the great work!

radu-gheorghe:
Awesome! Thanks for checking and merging!

PR #649, opened 2023-10-27
Make busybox securityContext configurable

### LINKED ISSUE #582: Run init container cp-solr-xml as nonRoot
Today we use busybox image with root user to run the cp-solr-xml init. This prevents `runAsNonRoot: true` to be set on the Solr pod. Either find a variant of busybox with non-root user, or explicitly specify `runAsUser: 65534` on the init container. Busybox image has a `nobody` user available ootb that I think we could use:

```bash
# docker run --rm busybox cat /etc/passwd
root:x:0:0:root:/root:/bin/sh
daemon:x:1:1:daemon:/usr/sbin:/bin/false
bin:x:2:2:bin:/bin:/bin/false
sys:x:3:3:sys:/dev:/bin/false
sync:x:4:100:sync:/bin:/bin/sync
mail:x:8:8:mail:/var/spool/mail:/bin/false
www-data:x:33:33:www-data:/var/www:/bin/false
operator:x:37:37:Operator:/var:/bin/false
nobody:x:65534:65534:nobody:/home:/bin/false
```

janhoy:
**Draft PR, only code, not docs, no helm support.**

The `busybox` official image runs as root in `cp-solr-xml` init-container, and there is no way to configure it otherwise, other than point to a different image that has been manipulated as non-root.

By adding a `SecurityContext` for the init container defaulting to the `nobody` user and setting `runAsNonRoot: true`, we have a good default. By also making the `securityContext` configurable, we allow for people to switch to a different image with other UID etc. Example:

```yaml
spec:
  busyBoxImage:
    tag: 1.36.1-glibc
  busyBoxSecurityContext:
    runAsUser: 123
```

Fixes #582


janhoy:
Alternatively, should perhaps the spec for `cp-solr-xml` init container be defined in yaml for the crd, e.g.:

```yaml
spec:
  cpSolrXmlInitContainer:
    image:
      registry: docker.io
      repository: busybox
      tag: 1.36.1-glibc
    imagePullPolicy: IfNotPresent
    securityContext:
      runAsUser: 65534
      runAsGroup: 65534
      runAsNonRoot: true
```

It would move some configuration from code out to values, which is always good. End users could then override only what they need, e.g. to use your own image with a different registry and user:

```yaml
spec:
  cpSolrXmlInitContainer:
    image:
      registry: public.ecr.aws
      repository: my-company/busybox
      tag: 1.37.0-custom
      imagePullSecret: foo
    securityContext:
      runAsUser: 1000
      runAsGroup: 1000
```

HoustonPutman:
@janhoy can we close this now that https://github.com/apache/solr-operator/pull/796 is merged? I think that is a superset of this functionality

PR #652, opened 2023-11-07
Update Solr version tag in local_tutorial.md

gerlowskija:
A contributor pointed out the tutorial no longer worked with operator 0.8.0 because some of the solr.xml changes made recently conflicted with the Solr version that the tutorial was using (8.3).

This commit updates the tutorial to use a Solr version within the range accepted by operator 0.8.0


gerlowskija:
Trying to refresh my understanding: once merged to `main`, this needs pulled into the `gh-pages` branch for the https://apache.github.io/solr-operator/docs/ site to be updated, right?

HoustonPutman (approved):
Nice! And yes, that is the correct way to backport this.

PR #694, opened 2024-03-25
Do not return metrics for liveness probe

### LINKED ISSUE #693: Liveness probe failing for Prometheus Exporter connected to a large SolrCloud
When running a Prometheus Exporter that is connected to a SolrCloud with many nodes/pods, the livenessProbe can fail with a `EOF` exception no matter what resources the pods are given.

I believe this is due to the kubelet not being able to handle a large amount of data being sent back in a probe response.

In order to fix this we should request back no metrics in the liveness probe. That way the response is empty.

HoustonPutman:
Fixes #693


HoustonPutman:
For some reason, in the 8.11 branch the `/-/healthy` endpoint still returns all metrics, so that doesn't fix this issue. Once we upgrade to 9x being the minimum supported version, then we can switch to `healthy` instead of just requesting no metrics.

gerlowskija (approved):
LGTM!

PR #706, opened 2024-05-25
Adding support for OpenShift securityContext

### LINKED ISSUE #466: SOLR StatefulSet gets a broken securityContext by default
Upon deploying a SolrCloud CR for the first time with the Operator (version 0.6.0), ZK is fine but the SOLR StatefulSet fails to start Pods because of:
```
create Pod demo-solrcloud-0 in StatefulSet demo-solrcloud failed error: pods "demo-solrcloud-0" is forbidden: unable to validate against any security context constraint: [provider restricted: .spec.securityContext.fsGroup: Invalid value: []int64{8983}: 8983 is not an allowed group provider restrictedfsgroup: .spec.securityContext.fsGroup: Invalid value: []int64{8983}: 8983 is not an allowed group]
```

If in the SolrCloud CR I specify an empty context, everything start fine:
```yaml
spec:
  customSolrKubeOptions:
    podOptions:
      podSecurityContext: {}
```

I don't know if this is expected, a bug or some edge case for OpenShift but there was no documentation for it.

computate:
Adding autodetection of solr-operator running on an OpenShift cluster to remove the default Solr fsGroup, and have an empty securityContext on OpenShift. Regular Kubernetes deployments have the same securityContext as before with fsGroup. 

I have successfully tested these changes in OpenShift Local and confirmed that an empty securityContext allows the solrcloud pod run on OpenShift. 

```yaml
spec:
  template:
    spec:
      securityContext: {}
```

Fixes #466

### DIFF
--- a/controllers/autodetect.go
+++ b/controllers/autodetect.go
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Copied from grafana-operator
+// With the Apache License: https://github.com/grafana/grafana-operator/blob/master/LICENSE
+// See: https://github.com/grafana/grafana-operator/blob/master/controllers/autodetect/main.go
+// Package autodetect is for auto-detecting traits from the environment (platform, APIs, ...).
+package controllers
+
+import (
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+)
+
+var _ AutoDetect = (*autoDetect)(nil)
+
+// AutoDetect provides an assortment of routines that auto-detect traits based on the runtime.
+type AutoDetect interface {
+	IsOpenshift() (bool, error)
+}
+
+type autoDetect struct {
+	dcl discovery.DiscoveryInterface
+}
+
+// New creates a new auto-detection worker, using the given client when talking to the current cluster.
+func NewAutodetect(restConfig *rest.Config) (AutoDetect, error) {
+	dcl, err := discovery.NewDiscoveryClientForConfig(restConfig)
+	if err != nil {
+		// it's pretty much impossible to get into this problem, as most of the
+		// code branches from the previous call just won't fail at all,
+		// but let's handle this error anyway...
+		return nil, err
+	}
+
+	return &autoDetect{
+		dcl: dcl,
+	}, nil
+}
+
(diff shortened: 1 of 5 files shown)


HoustonPutman:
I feel like it might be better to automatically tell if a cluster is openshift or not, and behave accordingly: https://github.com/grafana/grafana-operator/blob/master/controllers/autodetect/main.go

We can still let users over-ride the default, but just automatically choose the correct default for openshift. Thoughts?

gerlowskija:
Hey @computate - starting to follow up on some of these outstanding PRs - what do you think about Houston's suggestion above regarding auto-detection?  Is that something you'd be willing to add in here?

computate:
Thanks for following up @gerlowskija and @HoustonPutman , I am a big fan of Solr.  I would love to see the Zookeeper and Solr Operator work better on OpenShift, because I have had problems with it and had to switch to using bitnami helm charts instead. I haven't considered a solution like `if apiGroups[i].Name == "route.openshift.io" {` before. I will keep this update in mind, since this operator is also golang. I may need to create other issues for other things, I think I was also mostly blocked with issues with the Zookeeper Operator.

gerlowskija:
> I would love to see the Zookeeper and Solr Operator work better on OpenShift, because I have had problems

Sorry to hear it!  ZK Operator stuff we can't do as much about, but we'd love to have better support for OpenShift on the Solr side -  if you remember what the other issues were, please let us know with a ticket.  We're blind in some ways without that feedback!

gerlowskija:
Thinking a bit more on the proposal here, I'm a little leery about adding an explicit `openShift: true` flag.  One flag isn't all that bad, but there's a number of popular environments out there that might eventually want similar tweaking.

Auto-detection I feel much better about, since it needn't add to our CRD.  Could we _just_ do auto-detection to start, and revisit down the road if that ends up being insufficient?

computate:
Hi @gerlowskija and @HoustonPutman , I have now updated my branch to support the grafana-operator way of autodection of solr-operator running on OpenShift! If you would like to try this out, try my documentation for running OpenShift Local on your own computer with Windows 11 Pro, or Linux: 

https://smaq.computate.org/en-us/view/article/install-openshift-local-on-your-computer

Then 

```bash
install -d ~/.local/src/computate-gitops/
git clone https://github.com/computate-org/computate-gitops.git \
  ~/.local/src/computate-gitops/
cd ~/.local/src/computate-gitops/
oc apply -k openshift-local/gitops/base/
```

Then wait for the GitOps operator to become ready. You will be able to log into ArgoCD with your OpenShift Local kubeadmin credentials: https://openshift-gitops-server-openshift-gitops.apps-crc.testing/

```yaml
oc -n openshift-operators wait pod -l control-plane=gitops-operator \
  --for=condition=Ready --timeout=5m
oc -n openshift-gitops wait pod \
  -l app.kubernetes.io/name=openshift-gitops-application-controller \
  --for=condition=Ready --timeout=5m
```

Then deploy the zookeeper operator with GitOps: 

```bash
oc apply -k openshift-local/zookeeper-operator/applications/
oc apply -k openshift-local/zookeeper/applications/
```


Then deploy the solr operator with GitOps: 

```bash
oc apply -k openshift-local/solr-operator/applications/
oc apply -k openshift-local/solr/applications/
```

You should be able to find a SolrCloud running in the `solr` namespace. 

You will also find a log in the solr-operator pod: 

```logs
{"level":"info","ts":"2025-09-25T04:13:19Z","logger":"setup","msg":"autodetect","isOpenShift":true}
```

computate:
I have temporarily pointed my `solr-operator` ArgoCD Application to my branch until this gets merged: 
https://github.com/computate-org/computate-gitops/blob/main/openshift-local/solr-operator/applications/solr-operator/application.yaml#L13-L15

```yaml
    repoURL: https://github.com/computate/solr-operator.git
    targetRevision: openshift
    path: helm/solr-operator
```

Issue #650, opened 2023-10-30
Adding ObservedGeneration in the CRD


Shashankft9:
Currently, its hard to tell if solr cloud deployment has changed from "x" to "y" state, or if it has changed at all after I have done some changes in the CR.
Perhaps if we can add Observed Generator field and/or something like "LastTransitionTime", the CR could tell more about the current state of the solr cloud? This could be a good basis for a requirement we have, where we are tracking the transition of solr from "x" to "y".

Expected status:

```
status:
  backupRestoreReady: false
  internalCommonAddress: http://test-solrcloud-common.solr-cluster
  podSelector: solr-cloud=test,technology=solr-cloud
  readyReplicas: 3
  replicas: 3
  **observedGeneration: 1**
  solrNodes:
  - internalAddress: http://test-solrcloud-0.test-solrcloud-headless.solr-cluster:8983
    name: test-solrcloud-0
    nodeName: gccdelinttexw2
    ready: true
    specUpToDate: true
    version: 9.2.0
    **lastTransitionTime: 2023-10-27T12:53:56.989Z**
  - internalAddress: http://test-solrcloud-1.test-solrcloud-headless.solr-cluster:8983
    name: test-solrcloud-1
    nodeName: gccdelinttexw3
    ready: true
    specUpToDate: true
    version: 9.2.0
    **lastTransitionTime: 2023-10-27T12:54:56.989Z**
  - internalAddress: http://test-solrcloud-2.test-solrcloud-headless.solr-cluster:8983
    name: test-solrcloud-2
    nodeName: gccdelinttexw1
    ready: true
    specUpToDate: true
    version: 9.2.0
    **lastTransitionTime: 2023-10-27T12:55:56.989Z**
  upToDateNodes: 3
  version: 9.2.0
  zookeeperConnectionInfo:
    chroot: /test
    externalConnectionString: N/A
    internalConnectionString: test-solrcloud-zookeeper-0.test-solrcloud-zookeeper-headless.solr-cluster.svc.cluster.local:2181,test-solrcloud-zookeeper-1.test-solrcloud-zookeeper-headless.solr-cluster.svc.cluster.local:2181,test-solrcloud-zookeeper-2.test-solrcloud-zookeeper-headless.solr-cluster.svc.cluster.local:2181
```

gerlowskija:
[...]

PR #722, opened 2024-09-11
Added ObservedGeneration to solrcloud status

RavinaChidambaram:
Implemented #650 

Have added observed generation to the status of the solrcloud CR to track the transition of the solrcloud.

### DIFF
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -1160,6 +1160,9 @@ type SolrCloudStatus struct {
 	// BackupRepositoriesAvailable lists the backupRepositories specified in the SolrCloud and whether they are available across all Pods.
 	// +optional
 	BackupRepositoriesAvailable map[string]bool `json:"backupRepositoriesAvailable,omitempty"`
+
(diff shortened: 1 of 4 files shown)


gerlowskija (approved):
I left one in-line question, but otherwise this LGTM.  It seems to bring out status a little closer into line with best practices.

I'm ambivalent on the "lastUpdated" timestamp field that was described in the issue description.  It seems like there's less strong of a precedent around that field, and it's maybe more subject to some of the "false-signal" concerns I expressed in #650.  But I'm not against it - if others think it's worthwhile and are willing to add it, I'll hold off merging for a bit.  But otherwise I'll add a Chart.yaml change-entry and merge this sometime in the next few days!

Thanks @RavinaChidambaram !

RavinaChidambaram:
Thanks for the feedback and approval. I’m currently working on considering other scenarios as well around  `observedGeneration`, and I’ll address that within this PR. Please hold off on merging for a bit while I make these adjustments.

Also lets just address `observedGeneration` in this PR and then add `lastTransitionTime` in another PR

PR #729, opened 2024-11-15
fix: do not return metrics for prom-exporter probes re-fix (#694)

smoldenhauer-ish:
a re-fix for #694 as the used query parameter has to be 'name[]' instead of name**s**[]

according to the used HTTPServer code (0.16.0 in Solr 9 or 0.2.0 in Solr 8.11)
https://github.com/prometheus/client_java/blob/ed0d7ae3b57a3986f6531d1a37db031a331227e6/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java#L153


gerlowskija:
Thanks for the PR @smoldenhauer-ish .  Will start testing with an eye to merge soon.

gerlowskija:
Thanks @smoldenhauer-ish !

Issue #716, opened 2024-07-08
Unable to pass shareProcessNamespace to PodOptions


tommylogik:
I have a need to add `shareProcessNamespace` for use within a sidecar for some solr thread debugging and there doesn't seem to be a way to pass this option.

This should be able to be passed within the Solr Operator CRD.
```
apiVersion: solr.apache.org/v1beta1
kind: SolrCloud
metadata:
  name: explore
spec:
  customSolrKubeOptions:
    podOptions:
      shareProcessNamespace: true
```

Currently i have to use use kyverno to mutate the pod after the operator creates the statefulset and would love to be able to remove that destructive option for a more graceful approach in adding it to the helm chart values file.

gerlowskija:
Hey @tommylogik - sorry for the long delay.  I've [opened a PR](https://github.com/apache/solr-operator/pull/735) that should make this option available - would love a review or some manual testing if you're able to give it a try!

PR #735, opened 2024-12-12
Add shareProcessNamespace flag to podOptions

gerlowskija:
Boolean flag defaults unset, and is available in the CRD at `.spec.customSolrKubeOptions.podOptions.shareProcessNamespace`

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -139,7 +139,7 @@ type PodOptions struct {
 
 	// Should process namespace sharing be enabled on created pods
 	// +optional
-	ShareProcessNamespace *bool `json:"shareProcessNamespace,omitempty"`
+	ShareProcessNamespace bool `json:"shareProcessNamespace,omitempty"`
 
 	// Optional PodSpreadTopologyConstraints to use when scheduling pods.
 	// More information here: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -137,6 +137,10 @@ type PodOptions struct {
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	// Should process namespace sharing be enabled on created pods
+	// +optional
+	ShareProcessNamespace *bool `json:"shareProcessNamespace,omitempty"`
+
 	// Optional PodSpreadTopologyConstraints to use when scheduling pods.
 	// More information here: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
 	//
--- a/helm/solr/Chart.yaml
+++ b/helm/solr/Chart.yaml
@@ -42,12 +42,15 @@ annotations:
   # Allowed syntax is described at: https://artifacthub.io/docs/topics/annotations/helm/#example
   artifacthub.io/changes: |
     - kind: added
-      description: SolrClouds now support namespace sharing among pod containers in a pod.
+      description: Addition 1
       links:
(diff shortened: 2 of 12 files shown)


helm/solr/Chart.yaml
```diff
@@ -42,15 +42,12 @@ annotations:
   # Allowed syntax is described at: https://artifacthub.io/docs/topics/annotations/helm/#example
   artifacthub.io/changes: |
     - kind: added
-      description: Addition 1
+      description: SolrClouds now support namespace sharing among pod containers in a pod.
```
gerlowskija, line 45: Every time I come back to this project after a lull, I have trouble deciding which chart should hold a given change-entry.

Hopefully I've picked correctly here - I picked the 'solr' chart because the newly-exposed option affects the Solr pods (and not the operator pod itself)

gerlowskija:
This PR should allow us to close #716

helm/solr/Chart.yaml
HoustonPutman: Solr Operator, since that is the main chart (you can use the solr operator chart without using the solr chart, but not really the other way around). The Solr chart changeling is only used for changes relating to that helm chart.

api/v1beta1/common_types.go
```diff
@@ -137,6 +137,10 @@ type PodOptions struct {
 	// +optional
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 
+	// Should process namespace sharing be enabled on created pods
+	// +optional
+	ShareProcessNamespace *bool `json:"shareProcessNamespace,omitempty"`
```
HoustonPutman, line 142: Since the default is false, this probably doesn't need to be a pointer. pointer is only used when we want to differentiate the default value of the variable and the default when users leave the field empty.

HoustonPutman (commented):
this option should be mentioned in the Readme, `values.yaml` & `_custom_option_helpers.tpl` of the solr chart, since all the custom options are explicitly listed/set there.

helm/solr/Chart.yaml
gerlowskija: OK, thanks for clarifying; fixed!

api/v1beta1/common_types.go
gerlowskija: Makes sense; done.

gerlowskija:
> this option should be mentioned in the Readme, values.yaml & _custom_option_helpers.tpl of the solr chart, since all the custom options are explicitly listed/set there.

Oh, ok - I didn't realize we exposed every podOption there, but I guess that makes sense.  We still only want a Chart.yaml change-entry in the solr-operator helm chart I assume, per the in-line discussion above?

(Reminder to self: helm changes still need tested)

gerlowskija:
Alright, I think I addressed all your feedback @HoustonPutman , and I've tested out my changes to the 'solr' chart since my previous comment.  So this should be good to merge; I'll aim to do so shortly unless there's further feedback.

helm/solr/templates/_custom_option_helpers.tpl
```diff
@@ -36,6 +36,9 @@ resources:
 {{- if (include "solr.serviceAccountName.solr" .) -}}
 serviceAccountName: {{ include "solr.serviceAccountName.solr" . }}
 {{ end }}
+{{- if .Values.podOptions.shareProcessNamespace -}}
+shareProcessNamespace: {{ .Values.podOptions.shareProcessNamespace}}
```
HoustonPutman, line 40:
```suggestion
shareProcessNamespace: {{ .Values.podOptions.shareProcessNamespace }}
```

HoustonPutman (approved):

Issue #736, opened 2024-12-12
Mounting a volume to container is not clear


jamesla:
I am trying to attach a volume from a configmap to an init container but it is not entirely clear how to do it from the CRD spec. I've also been unable to find any examples of somebody doing this:

The code so far:

```yaml
---
apiVersion: solr.apache.org/v1beta1
kind: SolrCloud
metadata:
  name: test
spec:
  replicas: 1
  customSolrKubeOptions:
    podOptions:
      volumes:
        - source: 
            configMap:
          name: configset
          defaultContainerMount:
            mountPath: /config
            name: config.xml
      sidecarContainer:
        - name: config-loader
          image: alpine/curl:latest
          command:
            - "sh"
            - "-c"
            - "ls -la /config"
          volumeMounts:
            - name: configset
              mountPath: /config

---
kind: ConfigMap
apiVersion: v1
metadata:
  name: configset
data:
  config.xml: |
    <?xml version="1.0" encoding="UTF-8" ?>
     <config>
      ...
     </config>
```

It runs and the /config mount exists but it is always empty.

1. Are there any examples of volume mounts with this operator?
2. What am I doing wrong?

mcarroll1:
I've encountered a similar issue here https://github.com/apache/solr-operator/issues/675 (referencing an older issue here: https://github.com/apache/solr-operator/issues/484). 

I can't say exactly what's wrong here, but my guess would be that you need to add a line under `configMap` referencing the name (gleaning this from this doc: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap)

```
        - source: 
            configMap:
              name: configset
          name: configset
```

mcarroll1:
Turns out you can find the full spec here: https://github.com/apache/solr-operator/blob/main/config/crd/bases/solr.apache.org_solrclouds.yaml#L7038

mcarroll1:
Hi @jamesla . I've updated this PR to a configmap example directly: https://github.com/apache/solr-operator/pull/739. It appears to work for me locally. Let me know if that works for you.

PR #739, opened 2024-12-16
doc(volumes): add docs for attaching additional volumes

### LINKED ISSUE #675: Improve documentation for additional volumes
Hello,

I found it difficult to specify additional volumes in the `podOptions.volumes` values, specifically specifying a source for the volume. It appears @janhoy had a similar issue here: https://github.com/apache/solr-operator/issues/484. Reading that issue helped me determine the correct syntax, but it might be helpful to improve the documentation in this regard for posterity. The syntax seems to be specific to the `SolrCloud` resource, and not generic K8s.

I'm happy to submit documentation updates, but first: where is the best place to add such documentation? It might be too complex for the chart values doc. Perhaps it would be best in [these docs](https://apache.github.io/solr-operator/docs/solr-cloud/)? Another place?

Thanks!

mcarroll1:
Finally adding docs for this issue: https://github.com/apache/solr-operator/issues/675. There appears to be [another issue](https://github.com/apache/solr-operator/issues/736 ) on this topic, so it makes sense to add this to the CRD docs. First time adding docs here, so please review for style, etc.


janhoy:
Thanks for the contribution. I'm not sure if "Data storage" is the best place to describe custom volume mounts as they are NOT used to store index data but configuration or backup.

And `podOptions` is not something that is specific to Solr but to k8s, so perhaps mention in general without any examples that any standard pod option can be configured here, such as volumes.

When it comes to configuring Solr for backups, that is already covered by this documentation, for our managed backup feature, I think it will be confusing if we document a different approach as well: https://github.com/apache/solr-operator/tree/3cb870bd4d170390de829683ff4649f37e3a5658/docs/solr-backup

janhoy:
> These volume additions are actually part of the solrcloud CRD

The `customSolrKubeOptions.podOptions` that this PR is about is just passed on to k8s.

The link you just provided refers to the custom `backupRepositories` feature of the CRD, which offers to manage a backup volume for you. This is documented in the `SolrBackup` CRD docs.

I don't know where in the main `SolrClouds` doc to give more info about what you can do with `podOptions`. I think it should be enough to refer to k8s docs and mention that you can mount volumes etc. The default `values.yaml` also has some placeholder and comments as to what you can put below `podOptions`.

mcarroll1:
Sorry, I mean to link https://github.com/apache/solr-operator/blob/main/config/crd/bases/solr.apache.org_solrclouds.yaml#L7869, which appears to very closely mimic the backup volume CRD. 

To your point, perhaps this volume example is one specific slice of a very long spec that is mostly just passed onto K8s after some loose adaptation. However, I think it's still a problem that I find myself reading source code here to figure this out. Does the CRD itself produce any output that is accesible from our docs? It seems to be quite well commented.

gerlowskija:
I agree with janhoy that (as-is) this feels pretty similar to the existing [docs/solr-backup/README.md](https://github.com/apache/solr-operator/tree/3cb870bd4d170390de829683ff4649f37e3a5658/docs/solr-backup) content. 

But it'd be a great addition IMO if tweaked to overlap less with other backup docs (e.g. having the example us a config map, as suggested), and moved elsewhere.

To that end - @mcarroll1 you asked earlier about a place for extended docs/examples: we've actually got a good place for those!  What do you think about moving this to be a yaml file under [example/](https://github.com/apache/solr-operator/blob/main/example/).  YAML is a bit awkward maybe, but you can always put whatever explanatory prose you'd like as '#' comments.

Is that a reasonable path forward?

HoustonPutman:
So I like the additional example, but I also think the docs are useful. Maybe just don't include the example in the docs, and explain the top-level options, like the rest of the docs do.

Something like: (hopefully better, this is just a start)
```
These options can be found in `SolrCloud.spec. customSolrKubeOptions.podOptions.volumes`:

- `name`: A unique name for this volume, that will be used by volumeMounts in various containers
- `source`: This is a Kubernetes Volume reference. For example, a persistentVolume or a configMap
- `defaultContainerMount`: This is the specification for the volumeMount of this volume into the Solr container. Omit this if the volume should not be mounted in the Solr container.
```

Maybe you can link to the example as well. In the `source` explanation, we can probably link to Kubernetes documentation as well for all available options.

PR #702, opened 2024-05-06
Add security context config to SOLR containers

bentastic27:
fixes #489

Issue #489, opened 2022-11-03
Add support to set capabilities on Solr cloud container's security context  


pareekdevanshu:
Hi Team,
Currently It is possible to set capabilities for `initContainers` and `sidecarContainers`, Eg:
```
securityContext:
  capabilities:
    drop:
      - ALL
 ```
But, there is no way to set capabilities on Solr cloud container's security context, can you please help in adding support for it?

**Comment by ollixy** [comment]
@HoustonPutman is support for setting capabilities on the pod security context confirmed to be added eventually (and if it is, is there already an indication on when)? Or is this only under consideration and might be not added at all?

**Comment by mmoscher** [comment]
@ollixy you cannot set capabilities on the podSecurityContext, cf. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#podsecuritycontext-v1-core. What we need is a separate option to specify the securityContext of each container in the pod.

@HoustonPutman +1 for this issue, since solr is incompatible when using sokme restricted PSS and enforcing these with keyverno (for example).

I'll try to file an PR, but my knowledge about operators is quite limited.

**Comment by ollixy** [comment]
@mmoscher @HoustonPutman any news on this one?

**Comment by aaronsuns** [comment]
Any news on this one?

**Comment by AyzekTime** [comment]
Hi all! a new version was recently released (April 12, 2024, Apache Solr Operator™ v0.8.1 available, there is no solution to our problem) is there any news on our problem?

**Comment by janhoy** [comment]
> Hi all! a new version was recently released (April 12, 2024, Apache Solr Operator™ v0.8.1 available, there is no solution to our problem) is there any news on our problem?

@ollixy , @pareekdevanshu, @mmoscher, @aaronsuns , @AyzekTime 
This is an open source project, and we rely on contributions. If your day-job has a need for this and are willing to sponsor such a feature, then the best way forward is to arrange so that you can contribute a PR directly, and we'll help get it in to the next version.

**Comment by bentastic27** [comment]
I'm running into this issue on EKS.

@janhoy / @HoustonPutman Are these extra capabilties/privs ever actually needed? If not, we can likely hardcode in the changes made in this comment:
https://github.com/apache/solr-operator/issues/671#issuecomment-1875441585

Here is an example of a values.yaml I'm trying to use for a Solr Cluster:

```
securityContext: &securityContext
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["ALL"]
  runAsNonRoot: true
  seccompProfile:
[...]

PR #743, opened 2025-01-07
Allow users to configure Solr container's SecurityContext 

gerlowskija:
An alternate approach to #489.

#702 attempts to solve the same underlying issue by giving the Solr container in our STS template a hardcoded SecurityContext.  But that effort is, at the time of writing, stalled out: partly due to questions about whether the hardcoded SecurityContext would be too restrictive for some usecases, and partly due to an inability to test in the relevant environments.

This PR takes a different approach by leaving the securityContext unset by default, and instead giving interested users the ability to specify an arbitrary securityContext as a part of their SolrCloud (or as a 'solr' helm chart setting).  A securityContext can be provided in the resource YAML at `.spec.customSolrKubeOptions.podOptions.containerSecurityContext` (or using the `podOptions.containerSecurityContext` variable in the helm chart).

e.g.

```
apiVersion: solr.apache.org/v1beta1
kind: SolrCloud
metadata:
    ...
spec:
  solrImage:
    repository: solr
  customSolrKubeOptions:
    podOptions:
      containerSecurityContext:
        capabilities:
          drop:
            - ALL
        allowPrivilegeEscalation: false
  ...
  zookeeperRef:
    provided:
      chroot: "/"
      replicas: 3
      maxUnavailableReplicas: 1
```

### DIFF
--- a/controllers/solrcloud_controller_test.go
+++ b/controllers/solrcloud_controller_test.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	pointer "k8s.io/utils/pointer"
 	"strconv"
 	"strings"
 )
@@ -37,11 +38,6 @@ import (
 func newBoolPtr(value bool) *bool {
 	newBool := value
 	return &newBool
-}
-
-func newIntPtr(value int64) *int64 {
-	newInt := value
-	return &newInt
 }
 
 var _ = FDescribe("SolrCloud controller - General", func() {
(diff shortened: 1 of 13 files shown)


controllers/solrcloud_controller_test.go
```diff
@@ -38,6 +39,11 @@ func newBoolPtr(value bool) *bool {
 	return &newBool
 }
 
+func newIntPtr(value int64) *int64 {
```
HoustonPutman, line 42: You should be able to use `pointer.Int64(...)` for this. (The pointer library being "k8s.io/utils/pointer")

HoustonPutman: I see the I probably added the methods above that do the same thing. So we can always go and remove them independently, but let's not add another one.

HoustonPutman (approved):
Looks great! Just needs a changeling entry.

PR #749, opened 2025-01-21
Fix unset-var bug in 'verify_all.sh' script

gerlowskija:
A variable 'artifact' was being referenced outside of the for-in loop that created it.

### DIFF
--- a/hack/release/smoke_test/verify_all.sh
+++ b/hack/release/smoke_test/verify_all.sh
@@ -112,12 +112,11 @@ echo "Download all artifacts and verify signatures"
           gpg --verify "${artifact}.asc" "${artifact}" \
             || { echo "Invalid signature for ${artifact}. Aborting!"; exit 1; }
         fi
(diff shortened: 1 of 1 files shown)


janhoy:
Checked out the PR branch and re-ran the smoke tester. It now gets past the previous error!

janhoy (approved):

HoustonPutman:
Bizzare that this didn't happen before. I also ran into this when running the smoke tester. Let's merge this after the release. Unfortunately we can't commit to the `release-0.9` branch until after a success/failure of the RC.

janhoy:
> Bizzare that this didn't happen before. I also ran into this when running the smoke tester. Let's merge this after the release. Unfortunately we can't commit to the `release-0.9` branch until after a success/failure of the RC.

Ah, ok. So a workaround for those who want to vote is checking out your PR branch or applying https://patch-diff.githubusercontent.com/raw/apache/solr-operator/pull/749.patch

gerlowskija:
Thanks all for the review - sorry you hit this Jan!

PR #751, opened 2025-01-23
Avoid datetime deprecations in releaseWizard.py

gerlowskija:
Python 3.12 deprecates two popular datetime functions that are used throughout the Release Wizard, 'utcnow()' and 'utcfromtimestamp()', resulting in many distracting messages that make the other wizard output hard to read, e.g.

```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and
scheduled for removal in a future version. Use timezone-aware objects to
represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

This commit updates our 'datetime' usage in accordance with the Python recommendations.


HoustonPutman (approved):
gotta love python.

gerlowskija:
The releaseWizard requires at least Python 3.4 and the replacement functions ('now' and 'fromtimestamp') have been around at least that long, so this should be safe to merge in that respect at least.

HoustonPutman:
Nah, it's a good change. Thanks for doing the housekeeping!

janhoy (approved):

Issue #667, opened 2023-12-14
Ability to set custom hostname for SolrCloud


mcarroll1:
Hello,

Much like the `addressability.external.domainName` value in the SolrCloud resource, it would be helpful to have something like `addressability.external.hostName` to specify a particular hostname for external accessibility. It appears now that the hostname is always set to `<namespace>-<solrcloud-name>-solrcloud` (https://github.com/apache/solr-operator/blob/main/api/v1beta1/solrcloud_types.go#L1339) which is a bit redundant for my team's implementation of one SolrCloud per namespace (we end up with something like `solralpha-solralpha-solrcloud.domain.com`).

Thanks for your consideration!

janhoy:
Had the same need. Perhaps a configurable service-name template would be cool, where the default could be `serviceNameTemplate: %n-%r-solrcloud` given that %n=namespace, %r=releaseName.

HoustonPutman:
Yeah, I agree that could be cool!

PR #763, opened 2025-03-05
Make ingress hostname configurable

ps-muspelkat:
This PR addresses the possibility to add an `hostName` to the ingress configuration of a given solrcloud instance as described in #667.

The change is quit simple. Unfortunately, the `make prepare` step failed for me at the `fetch-licenses-list` step, so the `dependency_licenses.csv` stays empty in my case.

```
make: *** [fetch-licenses-list] Error 1
```

To make the e2e-tests start / work, I also needed to change the `RAW_GINKGO` command. Not sure if that is an issue only on my machine.

Happy for any feedback!


mikeywuu:
Sorry, had to re-create the PR. #764

PR #766, opened 2025-03-13
Fix upgrade to v0.9 by updating pod readiness gates

### LINKED ISSUE #761: Solr 9.8.0 pods do not become ready
We upgraded out test env from solr-operator 0.8.1 to 0.9.0 and from solr 9.7.0 to 9.8.0.

There seems to be an issue with the solr pods becoming healthy.

When we describe the pod , something

> status:
  phase: Running
  conditions:
    - type: solr.apache.org/isNotStopped
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: null
    - type: PodReadyToStartContainers
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2025-02-25T05:08:24Z'
    - type: Initialized
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2025-02-25T05:08:40Z'
    - type: Ready
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2025-02-25T06:10:51Z'
    - type: ContainersReady
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2025-02-25T05:09:42Z'
    - type: PodScheduled
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2025-02-25T05:08:10Z'

Notice that the "lastProbeTime" is Null.

It seems if we add to the "spec"

>   readinessGates:
    - conditionType: solr.apache.org/isNotStopped

Then the pods will start and will be working as expected.

What can i do without "manually" configure the readinessGates?

HoustonPutman:
Fixes #761 

The new feature forgot to add an update to the podTemplate for the new field being used.

### DIFF
--- a/controllers/util/common.go
+++ b/controllers/util/common.go
@@ -516,6 +516,12 @@ func CopyPodTemplates(from, to *corev1.PodTemplateSpec, basePath string, logger 
 		to.Spec.TopologySpreadConstraints = from.Spec.TopologySpreadConstraints
 	}
 
+	if !DeepEqualWithNils(to.Spec.ReadinessGates, from.Spec.ReadinessGates) {
+		requireUpdate = true
+		logger.Info("Update required because field changed", "field", basePath+"Spec.ReadinessGates", "from", to.Spec.ReadinessGates, "to", from.Spec.ReadinessGates)
+		to.Spec.ReadinessGates = from.Spec.ReadinessGates
+	}
+
 	return requireUpdate
 }
 


gerlowskija (approved):
This LGTM in terms of the production code, but should there be some test change accompanying this?

HoustonPutman:
Yeah, generally we test these methods in the unit tests via changing the customSolrKubeOptions. Unfortunately here, we do not support customizing this field, so we would need to change from one data type to another in order to test it... Doable in theory

PR #767, opened 2025-03-13
Prevent RAW_GINKGO unset error in e2e tests

gerlowskija:
Not all environments set particular `RAW_GINKGO` args, causing `make e2e-tests` to fail with an "unset variable" related error.

This commit tweaks our `manage_e2e_tests.sh` script to guard against this case.


gerlowskija:
I swear a different dev brought this up at some point in either an Issue or PR (though I can't find that comment now).  In any case, I've hit this a number of times across different machines, so seems worth fixing.

HoustonPutman (approved):

PR #769, opened 2025-03-14
Fix 'setup-zk' bash syntax error

### LINKED ISSUE #759: Zk-setup error
https://github.com/apache/solr-operator/blob/0ffddb6a4be25ffb3fdad6cb3ffe759d89c5fcf2/controllers/util/solr_security_util.go#L243

Zk setup container produce this error ```grep: ]: No such file or directory``` when authentication is activated.
I face this error after upgrading from 0.8.1 to 0.9.0 operator version.

gerlowskija:
Mismatched brackets were causing the 'grep' portion of the conditional to report an error when triggered.

Fixes #759

### DIFF
--- a/controllers/solrcloud_controller_basic_auth_test.go
+++ b/controllers/solrcloud_controller_basic_auth_test.go
@@ -353,7 +353,7 @@ func expectPutSecurityJsonInZkCmd(g Gomega, expInitContainer *corev1.Container) 
 	g.Expect(expInitContainer).To(Not(BeNil()), "Didn't find the setup-zk InitContainer in the sts!")
 	expCmd := "solr zk cp zk:/security.json /tmp/current_security.json -z $ZK_HOST >/dev/null 2>&1;  " +
 		"GET_CURRENT_SECURITY_JSON_EXIT_CODE=$?; if [ ${GET_CURRENT_SECURITY_JSON_EXIT_CODE} -eq 0 ]; then " +
-		"if [ ! -s /tmp/current_security.json ] || grep -q '^{}$' /tmp/current_security.json ]; then  " +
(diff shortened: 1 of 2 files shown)


HoustonPutman (approved):
Can we test this? I ask because your testing for the other ticket didn't see an error here...

gerlowskija:
I think our existing testing (see `tests/e2e/solrcloud_security_json_test.go`) is sufficient here.  It never caught the #759 error, because it's not actually an error ("just" some noise in the logs)!

`grep -q` does print a message to stderr, but it doesn't impact the return status of the `grep` call, which returns zero/non-zero based on whether it finds a match in `/tmp/current_security.json`.  `grep` doesn't care that you also asked it to search other files (i.e. `]`) that don't exist.  So despite the syntax error, the overall conditional still operated as we wanted it to.

We could add an additional test case to `solrcloud_security_json_test.go`, to cover the case of a security.json file whose contents are `{}`, but IMO that edgecase is rare enough that it's not worth bloating the runtime of our e2e tests.

If anyone disagrees though, lmk and I'll add it in.  Otherwise, I'll probably target early next week to merge this.

madrob:
I would consider pulling the entire scriptlets into separate files and then you can run spellcheck on those?

gerlowskija:
"Spellcheck" or "[ShellCheck](https://github.com/koalaman/shellcheck)"?

Assuming you mean "ShellCheck", but I'm not too familiar with it.  Seems like a good thing to have analysis on generally - our unit tests validate that the container "command" looks the way we expect, but nothing ever really ensures that it's even valid bash 😦 

Thanks for the pointer; will take a look and file a follow-on ticket if it seems promising!

madrob:
Yes, shellcheck. Auto-correct got me!

PR #770, opened 2025-03-17
Add default permission for replica balancing

### LINKED ISSUE #653: Getting unauthorized requests for cluster/replicas/balance in v0.8.0
Hi Team,

We've recently changed the operator version from v0.7.0 to v0.8.0 for our SolrCloud cluster (9.4). 
It seems that the new version requests the status of an async task named `balance-replicas-ScaleUp` every 60 seconds. Once it doesn't find one running, it sends a "replica/balance" request to the target Solr, therefore we're getting the following error for every attempt:

```
Recieved bad response code of 403 from solr with response: {
\"servlet\":\"default\",
\"message\":\"Unauthorized request, Response code: 403\",
\"url\":\"/solr/____v2/cluster/replicas/balance\",
\"status\":\"403\"
}
```

I found that the relevant metadata for this use case is written into the annotation field of the statefulsets object: `"solr.apache.org/clusterOpsLock": "{\"operation\":\"BalanceReplicas\",\"lastStartTime\":\"2023-11-09T14:28:18Z\",\"metadata\":\"ScaleUp\"}"`  

We're using the default security.json credentials and I'm not certain if there's anything to be changed in our settings for this matter.

Thanks!

HoustonPutman:
Fixes #653 

Will test this out tomorrow.


HoustonPutman:
Currently the v2 APIs do not support the named security rules (as of Solr 9.8.1) So we need to include the v2 path and the named security group for future and back compatibility.

PR #779, opened 2025-03-26
Configure operator logging in helm chart

### LINKED ISSUE #778: Allow easy configuration of logging in helm
Currently solr-operator logs by default in a console text format. Customers using various log aggregators request JSON format logging for more services.

By customizing command-line one can tweak the logger:

```
  -zap-devel
        Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error) (default true)
  -zap-encoder value
        Zap log encoding (one of 'json' or 'console')
  -zap-log-level value
        Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
  -zap-stacktrace-level value
        Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
```

While it is easy to set log level on the command line and thus also with the docker image using 'CMD' override, with the helm chart it is not that easy. Enable a new `logger` value section that allows user to configure each of the options listed above.

janhoy:
Adds this to `values.yaml`:

```
# Development mode configures certain settings for convenient development.
# When 'true', logging will use: encoder=console, level=debug, stacktrace-level=warn
development: false

# These are default logger settings in production mode, but can be overridden here
logger: {}
  #encoder: json
  #level: error
  #stacktraceLevel: error
```

Docs:
Skjermbilde 2025-04-03 kl  19 23 11

**Note** that today zap "Development" mode has been the default, but this PR changes that to `false`, triggering production mode by default, and thus also log-level INFO and encoding to JSON. This is perhaps a bit much change in a minor version even if it is easy for users to tailor it. At the same time it feels wrong that development is the default mode. We could keep the default of development=true and change it to false in v1.0?

Fixes #778

### DIFF
--- a/helm/solr-operator/Chart.yaml
+++ b/helm/solr-operator/Chart.yaml
@@ -60,8 +60,13 @@ annotations:
       links:
         - name: Github PR
           url: https://github.com/apache/solr-operator/pull/774
+    - kind: changed
+      description: The default logging format is now `json`, and default log level is `info`.
+      links:
+        - name: Github PR
+          url: https://github.com/apache/solr-operator/pull/779
     - kind: added
-      description: Ability to configure log encoder, level and stacktrace level
+      description: Ability to configure `development` mode as well as log encoder, log level and stacktrace log level
       links:
         - name: Github Issue
           url: https://github.com/apache/solr-operator/issues/778
--- a/helm/solr-operator/Chart.yaml
+++ b/helm/solr-operator/Chart.yaml
@@ -56,6 +56,11 @@ annotations:
   # 'kind' accepts values: "added", "changed", "deprecated", "removed", "fixed" and "security"
   artifacthub.io/changes: |
     - kind: added
+      description: Ability to configure log encoder, level and stacktrace level
+      links:
+        - name: Github Issue
+          url: https://github.com/apache/solr-operator/issues/778
+    - kind: added
       description: Addition 1
       links:
         - name: Github Issue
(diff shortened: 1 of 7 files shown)


HoustonPutman:
There is no impetus for a v1.0 right now, so I don't think there would be an issue doing this in v0.10.0

helm/solr-operator/Chart.yaml
```diff
@@ -55,6 +55,11 @@ annotations:
   # Allowed syntax is described at: https://artifacthub.io/docs/topics/annotations/helm/#example
   # 'kind' accepts values: "added", "changed", "deprecated", "removed", "fixed" and "security"
   artifacthub.io/changes: |
+    - kind: added
+      description: Ability to configure log encoder, level and stacktrace level
+      links:
+        - name: Github Issue
+          url: https://github.com/apache/solr-operator/issues/778
```
gerlowskija, line 62:
```suggestion
      links:
        - name: Github Issue
          url: https://github.com/apache/solr-operator/issues/778
        - name: Github PR
          url: https://github.com/apache/solr-operator/pull/779
```

helm/solr-operator/README.md
```diff
@@ -191,6 +191,10 @@ The command removes all the Kubernetes components associated with the chart and
 | tolerations | []object |  | Specify a list of Kubernetes tolerations for the Solr Operator pod |
 | priorityClassName | string | `""` | Give a priorityClassName for the Solr Operator pod |
 | sidecarContainers | []object |  | An optional list of additional containers to run along side the Solr Operator in its pod |
+| logger.devel | boolean | `false` | Enable development mode for the Solr Operator logger. This will log at the debug level with the console encoder. Default is `false` which logs at `info` level with `json` encoder. |
```
gerlowskija, line 194: [0] I could see the dev/prod distinction being useful for setting defaults totally unrelated to logging (docker image tag, resource limits, etc.).  Wdyt about moving this "mode" flag up to the top level, so it's easier to expand later to cover other things if desired?

janhoy:
I added some docs
* CLI opts
* Helm change-entry for the change of default log format and level
* upgrade-notes.md

gerlowskija (approved):
LGTM!

Issue #718, opened 2024-08-15
Cannot specify an imagePullSecret for solr-operator Helm chart


jstaf:
The solr-operator helm chart is missing the ability to specify a custom imagePullSecret to pull images from self-hosted container repos (that require authentication).

gerlowskija:
Hi @jstaf thanks for pointing this out - took a crack at it with [the PR here](https://github.com/apache/solr-operator/pull/734) - if you're willing to test it out or review I'd love a double-check!

PR #782, opened 2025-04-12
fix(deployment.yaml): correct the placement and syntax of the imagePu…

fruit-ninja:
- Provide a fix for the imagePullSecret value that allows Solr Operator users to host the image in their own repositories, protected by authentication. This fix will continue to accept an imagePullSecret to facilitate these use cases but in the correct format in the deployment.yaml
- I tested the original enhancement #734 from @gerlowskija (Thank you!!!) to add in the request for issue #718 but I was still getting an error where it was not using my secret for authentication. 
- I found the correct format from the [official Kubernetes docs](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret) and this [Stack Overflow solution](https://stackoverflow.com/questions/52192682/imagepullsecrets-not-working-with-kind-deployment). The imagePullSecrets should be placed in spec section.
- This fix has been tested to be working and correctly authenticating to my private repository.


fruit-ninja:
@HoustonPutman @gerlowskija would y'all please help me review this PR whenever you get an opportunity? Thank you!

HoustonPutman (approved):
Thanks so much for the fix here! Sorry it took such a ridiculously long time to get to it!

PR #785, opened 2025-05-02
Add chart README caveat for 'addtionalLibs'

gerlowskija:
I noticed recently that the setting wasn't working in a 'solrcloud', as the operator's main way to implement this feature is by creating `<sharedLib>` entries in the `solr.xml` file that it generates.  When users provide their own `solr.xml`, there's no way for the operator to inject/obey the "additionalLibs" setting.


gerlowskija:
The warning could use some word-smithing, and it probably makes sense to mention in the CRD field description too.  Just wanted to get some version of the PR up here before I forgot to document the limitation...

PR #804, opened 2025-10-31
Ensure initContainer sets SOLR_TOOL_OPTS based on SOLR_OPTS value

### LINKED ISSUE #717: It is impossible to setup TLS between Solr and Zookeeper
Tried setting up solr-operator with TLS between Solr and Zookeeper. (Zookeeper is setup with TLS, and Solr must connect to Zookeeper's TLS port.) It doesn't work or appear to have ever been tested.

Summary of problems I found before giving up:

* Solr cannot connect to Zookeepers using TLSv1.3 (will fail with `alert: protocol_version` until you downgrade).
* Contrary to the documentation, Solr's initContainers ignore `spec.solrZkOpts` completely... spec.solrZkOpts sets SOLR_OPTS, but the `solr zk` and `zkCli.sh` commands ignore SOLR_OPTS.
  * `solr zk` uses `SOLR_TOOL_OPTS` not `SOLR_OPTS` (and this is specific to Solr 9... Solr 8 uses a different env var) https://github.com/apache/solr/blob/main/solr/bin/solr#L601-L610
  * `zkcli.sh` uses `ZKCLI_JVM_FLAGS` not `SOLR_OPTS`
* The Solr TLS keystores/truststores don't even get mounted into the initContainers (and there is no way to do so), so even if you were able to specify the right combination of flags to `solr zk` and `zkcli.sh` in the initContainers, you can't add the certificates to setup TLS to Zookeeper anyways.
* The environment variables for the location of the truststore/keystore/passwords/etc. are not mounted in the initContainer
* There's no documentation on how to setup TLS between Solr and Zookeeper

To reproduce the issue:

Setup Zookeeper with TLS using the [Bitnami Helm chart](https://github.com/bitnami/charts/tree/main/bitnami/zookeeper) `helm install zookeeper bitnami/zookeeper -f zookeeper.yml`. Example `zookeeper.yml` (uses TLS for encryption, but not for authentication):

```yml
# note: solr flat-out cannot connect to zookeepers that use TLSv1.3,
# so we downgrade to TLSv1.2 here
jvmFlags: -Dzookeeper.ssl.protocol=TLSv1.2

# disabled for simplicity here
auth:
  client:
    enabled: false
  quorum:
    enabled: false

networkPolicy:
  extraIngress:
    - ports:
        - port: 3181
service:
  # zookeeper healthchecks and stuff are broken without this
  disableBaseClientPort: true
  headless:
    publishNotReadyAddresses: false
tls:
  client:
    # clients encrypt connections with TLS,
    # but do not use mTLS authentication
    enabled: true
    autoGenerated: true
    auth: "none"
  quorum:
    # use full mTLS auth+encryption between servers
    enabled: true
    autoGenerated: true
    auth: "need"
```

Then install solr-operator, and create a SolrCloud CRD. In this case I've generated some Certificate resources via cert-manger that the SolrCloud is using... 

```yml
apiVersion: solr.apache.org/v1beta1
kind: SolrCloud
metadata:
  name: solr
spec:
  replicas: 3
  solrImage:
    tag: 9.6.1
  dataStorage:
    persistent:
      reclaimPolicy: Delete
      pvcTemplate:
        spec:
          resources:
            requests:
              storage: "20Gi"
  solrTLS:
    # can't supply just a trustStore on its own, which is annoying
    # (maybe users want to get Solr -> Zookeepr TLS setup first before doing TLS for Solr itself?)
    pkcs12Secret:
      name: solr-tls
      key: keystore.p12
    keyStorePasswordSecret:
      name: solr-tls-password
      key: password
    # doesn't even get mounted into the initcontainer
    trustStoreSecret:
      name: solr-tls-password
      key: truststore.p12
    trustStorePasswordSecret:
      name: solr-tls-password
      key: password
  solrZkOpts: >
    -Dzookeeper.client.secure=true
    -Dzookeeper.clientCnxnSocket=org.apache.zookeeper.ClientCnxnSocketNetty
    -Dzookeeper.ssl.hostnameVerification=false
    -Dzookeeper.ssl.trustStore.location=/var/solr/tls/truststore.p12
    -Dzookeeper.ssl.trustStore.password=some-password
  zookeeperRef:
    connectionInfo:
      externalConnectionString: "zookeeper-0.zookeeper-headless.solr:3181,zookeeper-1.zookeeper-headless.solr:3181,zookeeper-2.zookeeper-headless.solr:3181"
      chroot: /solr
```

gerlowskija:
Resolves #717 by ensuring that all SOLR_OPTS values (which include all ZK connection settings) are also set in the SOLR_TOOL_OPTS variable utilized by `bin/solr zk`

### DIFF
--- a/helm/solr-operator/Chart.yaml
+++ b/helm/solr-operator/Chart.yaml
@@ -72,6 +72,13 @@ annotations:
           url: https://github.com/apache/solr-operator/issues/778
         - name: Github PR
           url: https://github.com/apache/solr-operator/pull/779
+    - kind: fixed
+      description: SOLR_TOOL_OPTS is now set, allowing 'setup-zk' initContainer to obey ZK connection settings
+      links:
+        - name: Github Issue
+          url: https://github.com/apache/solr-operator/issues/717
+        - name: Github PR
+          url: https://github.com/apache/solr-operator/pull/804
   artifacthub.io/images: |
     - name: solr-operator
       image: apache/solr-operator:v0.10.0-prerelease


gerlowskija:
Still need a changelog entry here, and a bit of cleanup for other env-vars we were setting that are unused.  (i.e. things that were utilized by zkcli.sh previously)

PR #805, opened 2025-11-03
Remove annoying whitespace

MMS-RBKN:
Please remove the annoying whitespace. We have an automation, that fails because of that.


HoustonPutman (approved):
Thanks for the fix, sorry for the issue with your deployments!

PR #807, opened 2025-11-04
Bump 'default' Solr version to 9.9.0

### LINKED ISSUE #806: Bump "default" Solr version to latest 9.x
I noticed recently that various parts of this repository default to using Solr 8.11, which is now EOL.  These usages range from the internal and relatively innocuous (e2e and "smoke test" scripting) to the user-facing and dangerous (documentation, example YAML, and the default tag for the Solr helm chart).

This ticket covers updating all of these places to use the latest Solr 9.x version instead (currently Solr 9.9.0).

There are a few places in the operator code itself that go out of their way to extend support to Solr 8.11.  These need cleaned up, but can be tackled in separate, follow-on tickets.

gerlowskija:
This leaves a few vistigial bits of operator code that exist in support of 8.11.  And this commit doesn't update our documentation to indicate that 8.11 is no longer supported.  Those can be handled in a subsequent commit or PR.

But otherwise this commit does everything necessary to make 9.9.0 the "default" and the version assumed by our docs.

Resolves #806

### DIFF
--- a/tests/scripts/manage_e2e_tests.sh
+++ b/tests/scripts/manage_e2e_tests.sh
@@ -76,7 +76,7 @@ if [[ -z "${KUBERNETES_VERSION:-}" ]]; then
   KUBERNETES_VERSION="v1.26.6"
 fi
 if [[ -z "${SOLR_IMAGE:-}" ]]; then
-  SOLR_IMAGE="${SOLR_VERSION:-9.8.1}"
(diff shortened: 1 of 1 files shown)


HoustonPutman:
Should we wait for 9.10?

gerlowskija:
Alright, I've updated this to use 9.10 now that it's out.  Making a few final edits and then merging...

PR #825, opened 2026-03-29
Fix two logic bugs reported in #822

### LINKED ISSUE #822: Two logic issues: service cleanup condition and balance replica condition
I noticed two small logic issues. 

## 1. `cleanupUnconfiguredServices()` condition looks inverted

Relevant code:

https://github.com/apache/solr-operator/blob/ca9d3c5c37a59f29570a6b49a8da5dc614aba75e/controllers/solrcloud_controller.go#L683-L716

```go
if pod.Annotations == nil && pod.Annotations[util.ServiceTypeAnnotation] != "" {
```

This condition looks inconsistent. If `pod.Annotations == nil`, then reading `pod.Annotations[...]` does not make sense. If `pod.Annotations != nil`, the left side is false, so the code falls into the else branch, but the else branch comment says it is for pods missing the annotation.

## 2. `BalanceReplicasForCluster()` comment and code do not match

Relevant code:

https://github.com/apache/solr-operator/blob/ca9d3c5c37a59f29570a6b49a8da5dc614aba75e/controllers/util/solr_scale_util.go#L35-L41

```go
// If the Cloud has 1 or zero pods, there is no reason to balance replicas.
if statefulSet.Spec.Replicas == nil || *statefulSet.Spec.Replicas < 1 {
    balanceComplete = true
}
```

The comment says “1 or zero pods”, but the condition only matches < 1, not <= 1.

oyjhl:
Fixes #822

### DIFF
--- a/controllers/solrcloud_controller.go
+++ b/controllers/solrcloud_controller.go
@@ -680,7 +680,7 @@ func (r *SolrCloudReconciler) cleanupUnconfiguredServices(ctx context.Context, s
 		return
 	}
 	for _, pod := range podList {
-		if pod.Annotations == nil && pod.Annotations[util.ServiceTypeAnnotation] != "" {
+		if pod.Annotations != nil && pod.Annotations[util.ServiceTypeAnnotation] != "" {
 			if onlyServiceTypeInUse != pod.Annotations[util.ServiceTypeAnnotation] {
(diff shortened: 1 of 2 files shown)


HoustonPutman (approved):

HoustonPutman:
Thanks for the fixes!

Issue #821, opened 2026-03-16
Clarify whether the supported Solr versions includes Solr 10


plumdog:
Solr 10 has been released. https://solr.apache.org/downloads.html#solr-1000

https://github.com/apache/solr-operator/blob/main/docs/upgrade-notes.md#solr-versions just shows "9.4+" and it isn't clear to me whether this just means ">=9.4, <10" or ">=9.4".

janhoy:
Feel free to test whether 10.0 works already. It may work partially, but obviously not support any new features. So perhaps some disclaimer about what works and not would be good to add. I have no idea as I have not tested it yet.

HoustonPutman:
Yeah, I tried testing before the release but got caught and didn't have time to deep dive. We should probably put <10 until we can figure it out...

HoustonPutman:
Ok, I've gotten the E2E tests to work on my laptop again, and it's not looking good for Solr 10 or 9.10 even...

I'll do some more work to isolate working versions and update our docs. Then we can start to fix things.

HoustonPutman:
Ok from my findings:

- Solr 9.8 - 0 failures
- Solr 9.9 - 2 flaky tests: (both failed 2 out of 3 runs)
  - `Scale Up with replica migration using an ingress for node addresses`
  - `Scale Down without replica migration`
- Solr 9.10 - 0 failures
- Solr 10.0 - No tests pass

So I think we are good for the 9.x line at least for the things that are tested. Solr 10 does not work at all, and we need to fix the operator and/or solr. In the meantime, I'll make a PR for the docs.

epugh:
Let me know if some of the CLI changes made life hard...   I'm having a small bit of regret on some changes, like eliminating the `-c` option in Solr 10 versus just making it a no-op when I run `bin/solr start -c`.....

janhoy:
Yea, making `-c` a no-op would be smoother. But too late for 10.0.0.

HoustonPutman:
We could just support 10.1.0, and have that fixed there. I'm not sure how we would support 9.x and 10.x at the same time if we can't pass `-c` to both.

janhoy:
@epugh WDYT about bringing `-c` option back again as a NOP for 10.1?

epugh:
> [@epugh](https://github.com/epugh) WDYT about bringing `-c` option back again as a NOP for 10.1?

I think that would be best.   I've bumped my head on this as well.  Want to toss me a JIRA and i'll do the work?

[...]

PR #826, opened 2026-03-30
Solr 10 support

tomaszpolachowski:
It refers to issue #821

There 2 issue around running Solr Operator againt Solr 10:
* `host` parameter deprecated completely - must be solr.host.advertise instead
* Number of items in default solr.xml are no longer valid - rewritten

### DIFF
--- a/controllers/util/solr_util.go
+++ b/controllers/util/solr_util.go
@@ -378,6 +378,10 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 			Value: solrHostName,
 		},
 		{
+			Name:  "SOLR_HOST_ADVERTISE",
+			Value: solrHostName,
+		},
+		{
 			Name:  "SOLR_LOG_LEVEL",
 			Value: solrCloud.Spec.SolrLogLevel,
 		},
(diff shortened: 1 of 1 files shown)


controllers/util/solr_util.go
```diff
@@ -377,6 +377,10 @@ func GenerateStatefulSet(solrCloud *solr.SolrCloud, solrCloudStatus *solr.SolrCl
 			Name:  "SOLR_HOST",
```
HoustonPutman, line 377: Can we get rid of this now? Since we have `SOLR_HOST_ADVERTISE`?  Or will it be used by the CLI when we interact with it via other commands?

HoustonPutman (commented):
Looks good, we will have to ensure the integration tests look good.

controllers/util/solr_util.go
tomaszpolachowski: It's for backward compatibility only. Depends if support for versions before introducing `solr.host.advertise` is necessary. If not then along with removal some tests and `probeCommand` in solr_security_util.go would require update.

janhoy:
I tagged this with the 1.0.0 milestone. Will try to add a few other issues that are needed/wanted for next major..

klaus-tux:
Hi @HoustonPutman @janhoy ,
I wanted to ask politely, if there is a current estimation for the release date of the solr-operator 1.0.0 ? I'm looking forward to solr-operator supporting solr 10 :-)
Thanks in advance!
Best regards,
Klaus Sperner

HoustonPutman:
So unfortunately the main issue is that the command line options for Solr 9 and Solr 10 are not compatible when running Solr in cloud mode. So there's no way for us to natively support both at the same time. The way we planned on fixing this was releasing a Solr 10.1 that added compatibility for the `-c` flag used in all previous versions of Solr. However, the release of all new Solr versions have been blocked for now. We will get a new release of the Solr operator soon, which should do the necessary things to support Solr 10. But we also need that new Solr 10.1 release. The delay is very frustrating, and I empathize. Hopefully we can get that new Solr release out soon.

klaus-tux:
Hi @HoustonPutman , thank you for answering my question and for the transparency. Highly appreciated.

Issue #820, opened 2026-03-04
Support new metrics format for 10.0


janhoy:
In Solr 10.0, released today, the Prometheus exporter is gone. Instead each solr node exposes prometheus-format metrics individually on `/solr/admin/metrics`. 

Users on Solr 9.x should still be able to run PrometheusExporter as before which will work with Solr 9 but not solr 10.

Users moving to Solr 10, who want metrics, should upgrade to a new Solr Operator which supports configuring Prometheus to scrape solr endpoints intead of deploying solr-exporter.

The `SolrClouds` CRD and thus also the `solr` helm chart should make metrics exporting simple to use OOTB
* Default to open `/solr/admin/metrics` endpoints, document how to configure a BasicAuth user with access to this endpoint
* Add a new CRD key to enable prometheus ServiceMonitor pointing to the metrics endpoint, and make this key exposed as helm value
* Make the operator manage these ServiceMonitors

Users who want to push metrics to OTEL collector over OTLP can probably do this without custom operator support, by simply setting some environment variables to configure the OTEL SDK. Point users to relevant Solr documentation for this.

PR #827, opened 2026-04-27
Add Solr 10 compatibility support

danielmeint:
## Add Solr 10 compatibility support

Adds version-conditional Solr 10 support to the operator while preserving Solr 9.x behavior. All major operator-managed flows (bootstrap, scaling, security, TLS, backups, ingress, rolling upgrades) work against Solr 10.0.

Builds on #826 and addresses #821. Prometheus exporter is out of scope: `solr-exporter` was removed in Solr 10 (#820).

> [!WARNING]
> **Breaking change:** introduces a new field `spec.solrMajorVersion` (int, 8–10) that is **required at the CRD level**; CRD validation rejects CRs that omit it. Hand-authored CRs must add the field (e.g. `solrMajorVersion: 9`) before upgrading the operator. The `helm/solr` chart defaults the new `solrMajorVersion` value to `9` (matching the default `image.tag` of `9.10.0`), so existing chart users keep working without action; set it to `10` when pointing `image.tag` at a Solr 10.x image. An explicit field avoids implicit detection from image tags, which would be unreliable for custom/internal tags.

### What changed

Version detection is driven by `(*SolrCloud).IsSolr10OrLater()`, which reads `spec.solrMajorVersion`. No image-tag parsing.

| Area | Change |
|---|---|
| `solr.xml` | New `DefaultSolrXMLForSolr10` template, drops `genericCoreNodeNames`, `hostContext`, `allowPaths`, `metricsEnabled`; switches `host` to `${solr.host.advertise:}` |
| Host advertise | `SOLR_HOST_ADVERTISE` env var (replaces deprecated `host` sysprop) |
| Modules | `SOLR_MODULES` env var; `sharedLib` skips removed `/opt/solr/contrib/<module>/lib` and `/opt/solr/dist` |
| `hostPort` sysprop | Skipped on Solr 10 |
| TLS cluster prop | `setUrlSchemeClusterPropCmd` uses `solr zk cp` instead of removed `cloud-scripts/zkcli.sh` |
| Secure probes | `useSecureProbe` uses `solr api --solr-url …` instead of removed `-get` |
| E2E helpers | `callSolrApiInPod` uses `--solr-url` and `--credentials user:pass` (Solr 10 dropped `-Dbasicauth=…` via `JAVA_TOOL_OPTIONS`) |

### Test plan

- [x] Unit tests: `IsSolr10OrLater`, both `solr.xml` templates, both branches of `useSecureProbe`
- [x] E2E against Solr 10.0.0: Basic, Scale Down/Up (with and without replica migration), Security JSON (provided + bootstrapped), TLS Secrets (No Client TLS, Client TLS, ClientAuth Need/Want, CheckPeerName, VerifyClientHostname), TLS Mounted Dir (ClientAuth Want), Local-directory backups (recurring + single), Ingress, Managed Rolling Upgrades
- [x] Solr 9.8.0 Basic regression baseline
- [ ] Prometheus Exporter: fails as expected, exporter removed in Solr 10 (#820)

### Notes

Solr 10 logs a `"deprecated system property host"` warning at startup. After investigation this is a Solr-side false positive (`bin/solr` in 10 doesn't translate `SOLR_HOST` to `-Dhost`, and Solr 10's stock `solr.xml` ships with the same `<str name="host">${solr.host.advertise:}</str>`). Removing `SOLR_HOST` from the pod spec doesn't silence it. Left as-is.

### DIFF
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -1327,6 +1327,39 @@ func (zkInfo ZookeeperConnectionInfo) ZkConnectionString() string {
 	return zkInfo.InternalConnectionString + zkInfo.ChRoot
 }
 
+// SolrMajorVersion extracts the major version number from a Solr image tag.
+// Returns 0 if the tag cannot be parsed (e.g. "latest", "nightly", custom tags).
+func SolrMajorVersion(imageTag string) int {
+	tag := strings.TrimPrefix(imageTag, "v")
+	if idx := strings.Index(tag, "-"); idx >= 0 {
+		tag = tag[:idx]
+	}
+	major := tag
+	if idx := strings.Index(tag, "."); idx >= 0 {
+		major = tag[:idx]
+	}
+	v, err := strconv.Atoi(major)
+	if err != nil {
+		return 0
+	}
+	return v
+}
+
+// IsSolr10OrLater returns true if the given image tag represents Solr 10.0 or later.
+// Unparseable tags (e.g. "latest") are treated as pre-10 for backwards compatibility.
+func IsSolr10OrLater(imageTag string) bool {
+	return SolrMajorVersion(imageTag) >= 10
+}
+
+// IsSolr10OrLater returns true if this SolrCloud's image tag represents Solr 10.0 or later.
+// A nil SolrImage is treated as pre-10.
+func (sc *SolrCloud) IsSolr10OrLater() bool {
+	if sc.Spec.SolrImage == nil {
+		return false
+	}
+	return IsSolr10OrLater(sc.Spec.SolrImage.Tag)
+}
+
 // UsesHeadlessService returns whether the given solrCloud requires a headless service to be created for it.
 // solrCloud: SolrCloud instance
 func (sc *SolrCloud) UsesHeadlessService() bool {
--- a/api/v1beta1/solrcloud_types.go
+++ b/api/v1beta1/solrcloud_types.go
@@ -74,6 +74,15 @@ type SolrCloudSpec struct {
 	// +optional
 	SolrImage *ContainerImage `json:"solrImage,omitempty"`
 
+	// Solr major version (8, 9, or 10). Determines version-specific operator
+	// behavior such as solr.xml format, CLI flags, and env vars.
+	// Required — image tags alone are not reliably parseable.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=8
+	// +kubebuilder:validation:Maximum=10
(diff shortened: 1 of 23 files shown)


ArbitraryCritter:
This looks great, I'll see if I can give it a whirl in our test cluster.

ArbitraryCritter:
Just tested this on our 3 node staging cluster, which sees a reasonable test load, I'll keep it running in this configuration for ~a week and report any issues we may see.

Cluster state before start:
solr-operator: 0.9.1
solr: 9.10

Process:
- I first upgraded the solr-operator without any other modification (eg. helm chart and solr-operator image built from this branch).
- Upgraded the solr cluster to 10.0 based image (with some custom modules)

Findings:

There were no issues with upgrading the solr-operator itself, that just worked. Only thing to note is that the SolrMajorVersion seems to default to 9, which seems quite sane, but also slightly different from what is mentioned in the PR description, I'd expected it to complain that I needed to set a version to proceed.

As is noted in the issue, the prometheusExporter is gone from Solr 10 and upgrading with it will not work, it has to be disabled first. Failing to remove it caused the upgrade process to stall, requiring a rollback.

The SOLR_MODULES behavior seems a bit different with the new solr helm chart. We've been using a custom docker image that contains a solr module specific to our own environment (used for access checks). We've been overriding the SOLR_MODULES env directly in the dockerfile. Before the solr helm chart upgrade the SOLR_MODULES variables wasn't set by the helmchart, after the upgrade it was (with the value SOLR_MODULES=s3-repository).
This was easily remedied by adding the appropriate modules to "solrOptions.SolrModules", but it does seem that something about the SolrModules env behaves a bit differently in the new solr chart.
I don't think this is a bug, it's probably more us misusing the SOLR_MODULES concept by overriding it in the Dockerfile rather than the helm chart, but perhaps a note should be added to the release that users should be sure to define the solr modules they use through the helm chart (or something similar).

All in all this seems to work as expected.

danielmeint:
Awesome, thanks a lot for the thorough report, I really appreciate it.

Fair point on the discrepancy regarding the `solrMajorVersion` param. To clarify, the _CRD field_ is currently required, but the _helm chart_ defaults to 9. I don't have a strong intuition on whether that's the best way to handle it, to be honest. I'll at least update the PR description and upgrade notes to make this more explicit.

Same for disabling the prometheus exporter and the behavior change regarding `SOLR_MODULES`, I'll add more info to the upgrade notes to explain this better.

Again, thank you for taking the time 🙂

coolstim:
how does this compare to https://github.com/apache/solr-operator/pull/826 
introducing a breaking change does not feel right, i was able to get the current operator to work against solr 10 just by pointing it to a valid solr.xml

@janhoy thoughts on this?

PR #832, opened 2026-06-08
Support global.imagePullSecrets in both helm charts

### LINKED ISSUE #338: Use global.imagePullSecrets at solr-operator helm chart

mattiasbpersson:
## Description

Adds `global.imagePullSecrets` support to the solr-operator helm chart deployment.

The existing `image.imagePullSecret` field is rendered inside `containers[]` in the deployment template, which is not a valid Kubernetes pod spec field. Kubernetes silently ignores it, so pull secrets are never applied to the operator pod. This makes it impossible to deploy the solr-operator from a private registry.

This PR adds `global.imagePullSecrets` at the pod spec level (`spec.template.spec.imagePullSecrets`), consistent with how the solr helm chart already handles it.

Supports both formats:
- Kubernetes-native: `[{name: "my-secret"}]`
- Plain strings: `["my-secret"]`

## Changes

- `values.yaml` — Added `global.imagePullSecrets` default
- `templates/deployment.yaml` — Render `imagePullSecrets` at pod spec level
- `Chart.yaml` — Changelog entry
- `README.md` — Documentation entry

## Related

- Closes #338
- Supersedes #343 (stale since 2021, original author inactive)


HoustonPutman:
This is great, thanks for starting this off! I went ahead, and added the same for the Solr helm chart, and both merge the local and global imagePullSecret options, so that either can be used.

PR #834, opened 2026-06-08
Migrate documentation to Antora

### LINKED ISSUE #833: Solr Operator documentation should live in the Solr Ref guide under its own section
The Solr Ref Guide is built using Antora, which allows for different components with different versioning schemes to live in the same documentation website. We can have both Solr and the Solr Operator documentation live there, while having their own versions and release cycles.

HoustonPutman:
Part of #833 

`make docs` will build a local version of the site (without Solr).
`make check-docs` will validate everything. This is now included in the linting process.


anshumg (approved):
Sanity checked and it looks good to me. Certainly didn't read it all :) 
Thanks @HoustonPutman !

docs/modules/getting-started/pages/local-tutorial.adoc
```diff
@@ -1,41 +1,41 @@
...
-
-    Unless required by applicable law or agreed to in writing, software
-    distributed under the License is distributed on an "AS IS" BASIS,
-    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-    See the License for the specific language governing permissions and
-    limitations under the License.
- -->
-
-# Solr on Kubernetes on local Mac
+= Solr on Kubernetes on local Mac
```
anshumg, line 1: Should the license header be before the first line?

HoustonPutman: Apparently not, the solr ref guide is the same way

PR #836, opened 2026-06-11
Emit Kubernetes events from the operator controllers

### LINKED ISSUE #120: Add events to the controllers
The only way to get information from the controllers is currently via the logs. Kubebuilder has a way of adding events to the resource, so that kubectl describe will show what is happening.

We should add these events for all controllers in the Solr Operator.

https://book-v1.book.kubebuilder.io/beyond_basics/creating_events.html

HoustonPutman:
Closes #120 

Fills in the `// TODO: Create event` markers and adds event recording across the SolrCloud, SolrPrometheusExporter, and SolrBackup controllers (config validation, scheduled restarts, cluster operations, replica migration, PVC expansion, scaling, and backup lifecycle).

Also grants the operator RBAC to create events — without it the broadcaster silently dropped every event, so the previously merged PVC-expansion events never  persisted either.

Adds a no-op `EventRecorder` so reconcilers call the recorder unconditionally, and wires recorders for the exporter and backup controllers.

Tested at three layers: unit (`FakeRecorder`), envtest reconcile, and e2e (backups, scaling, rolling upgrades).

A lot of work done by Claude on this.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -42,6 +42,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - events
+  verbs:
+  - create
+  - patch
+- apiGroups:
+  - ""
+  resources:
   - persistentvolumeclaims
   verbs:
   - delete
--- a/controllers/event_recorder.go
+++ b/controllers/event_recorder.go
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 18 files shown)


janhoy:
Cool, so this means that if you monitor kube events, you won't have to go into the solrOperator logs to see what operations were done or failed anymore?

I cannot contribute with much more than clicking the "Copilot" review button right now :)

HoustonPutman:
> Cool, so this means that if you monitor kube events, you won't have to go into the solrOperator logs to see what operations were done or failed anymore?

Yeah, a lot of information can be populated there. Errors plus operational things like restarting pods, backups, etc. So the logs should be purely for debugging now!

HoustonPutman:
> I cannot contribute with much more than clicking the "Copilot" review button right now :)

Looks like co-pilot likes it. Luckily this is pretty low-risk since we are just adding events in addition to existing logging, and the e2e tests make sure that it doesn't affect runtime.

janhoy (approved):

PR #840, opened 2026-07-10
fix: treat NotFound from pod delete in rolling update as success

AruneshDwivedi:
DeletePodForUpdate deletes the target pod with a UID precondition, but any error from that delete (including Already-NotFound) was logged and surfaced as a PodUpdateError warning. When the pod is already gone - the desired postcondition for this update path - the StatefulSet recreates it, so the NotFound is expected. Now apierrors.IsNotFound is checked and treated as success (normal PodUpdate event) while other delete errors still warn. Added two regression tests (NotFound -> normal event, real error -> warning). Verified with go build, go vet, and the targeted tests pass; the package's Ginkgo/ envtest BeforeSuite is environment-dependent and not related to this change.

### DIFF
--- a/controllers/events_test.go
+++ b/controllers/events_test.go
@@ -28,9 +28,11 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


HoustonPutman (approved):
Good call, this code hasn't been released, so good find!

